PR #6, opened 2016-12-20
Worked on RPC Applet and ReadMe.md

JSmith-BitFlipper:
I moved the rpc applet to use the QT libraries, which are a bit more mature than GTK.

I also polished up the README.md files.


adiabat:
Awesome, the python qt interface works on my machine!

I will have to learn about this qt stuff :)

PR #8, opened 2016-12-29
Implemented an RPC Console into the QT Client

JSmith-BitFlipper:
Users can type RPC commands directly into the QT client and get output from the lightning server.


adiabat:
lots has changed; closing now but at QT client would be great; starting work on a GTK attempt

PR #10, opened 2017-01-26
Add lit home directory to store files such as headers.bin

takaya-imai:
This PR is to create lit home directory.
Default directory is `$HOME/.lit` and can be changed by `lit` command option `-dir`.

### DIFF
--- a/lit.go
+++ b/lit.go
@@ -4,6 +4,8 @@ import (
 	"flag"
 	"fmt"
 	"log"
+	"os"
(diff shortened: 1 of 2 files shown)


adiabat:
Looks good, thanks a lot!
For the previous usage, you can say `./lit -dir .` and it will put files in the same directory for testing.  I'm not sure how it will work on windows / mac but can check out compatibility later.

PR #12, opened 2017-02-05
Add unit-tests for msglib.go and stdlib.go in lnutil

takaya-imai:


adiabat:
Thanks for the test coverage!  I wish we didn't have to write these functions in go, but we do.  So, better to test and make sure it works!

PR #13, opened 2017-02-05
add an explanation of dir option for Bob

takaya-imai:
I forgot to fix walkthrough doc when I sent PR #10 .


kallewoof:
Instead of opening a new PR, you can update your existing one by adding a commit to it. (Just commit to the branch.)

You can also squash commits together into one commit if that is appropriate. The thoughts on when to squash and such vary between people, though so it's not always a good idea.

takaya-imai:
@kallewoof thanks. I squash the commits into one and push it.

PR #14, opened 2017-02-05
Add readline support to lit-af

kallewoof:
Uses https://github.com/chzyer/readline -- not sure if it's the best option as I didn't do much research but it may be good for now.

### DIFF
--- a/cmd/lit-af/autocomplete.go
+++ b/cmd/lit-af/autocomplete.go
@@ -1,0 +1,96 @@
+package main
+
+import (
+	"fmt"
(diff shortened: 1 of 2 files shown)


kallewoof:
I started working on an autocompleter as well, to make this even better, but I realized I have no idea what some of these commands do. Fan and sweep in particular. I've pre-emptively added autocompletion to 'help' to take the command names, as I presume it will be possible to do "help <cmd>" at some point.

adiabat:
[TAB] FTW!

PR #18, opened 2017-02-08
Colorize methods and outputs.

kallewoof:
I may have gone a bit overboard, but time will tell. 

I am colorizing all satoshi outputs in the following manner:
- BTC values are white + bold,
- mBTC values are regular,
- <1mBTC values are grayish (faint).

It feels like it expresses amounts in a easy to comprehend manner, but I am open to changing. :)


adiabat:
Thanks!  This is super leet.  Maybe we can add a command line option to make everything green text for more 1337ness :)

PR #19, opened 2017-02-09
Renamed color vars to their purpose, rather than their colors.

kallewoof:
Also added some more colors (addresses & outpoints).


adiabat:
Cool.  I'm liking the new lit-af UI.
Actually I think the colors import will probably need to move to the cmd folder and just work with lit-af eventually, because lit itself will need to log to files instead of stdout, which don't really support color.  But can modify in lnutil and move later.

kallewoof:
How about incoming messages? Should they simply be uncolored?

PR #20, opened 2017-02-09
Add unit-test for btclib.go

takaya-imai:
This is an unit-test for lnutil/btclib.go.
There are some TODOs but I add them later.


adiabat:
Thanks!

PR #23, opened 2017-02-10
Trivial: Comment on wop.Hash being unnecessary in bloom filter generation.

kallewoof:
The wop.Hash addition to the local bloom filter is probably useless, so adding a comment about this. Someone else should verify whether this is the case and/or (preferably) make tests to ensure it and then remove this line.


adiabat:
Actually I think I will just remove this filter add command as we're now pretty sure it does nothing right?

kallewoof:
That works, but I'm not ultra confident it's actually the case. Remove and see if things work for awhile, if that is acceptable? :)

jnewbery:
Looks like this was fixed in 6a9fdb90. Suggest we close this issue (with thanks to @kallewoof!)

kallewoof:
Sounds good! (I forgot about this completely.)

PR #24, opened 2017-02-15
Add unit-test for curvelib.go

takaya-imai:
I found two panic points but I think its priority might be low.
If no needs to fix them, I delete TODOs for the points and push again.


adiabat:
Looks good, thanks!

PR #27, opened 2017-02-20
Add unit-test for keygen.go in portxo

takaya-imai:
I add unit-test for keygen.go.

- KeyGen.String() has a case with panic. If it is better to fix it, I fix it and push later.
- KeyGen.Bytes() returns a byte slice in case of KeyGen.Depth = 6 (more than 5). I am not sure that this behavior is suitable so I added test code as the same way as other tests.


adiabat:
Thanks!  Hope to expand / add integration tests soon

takaya-imai:
@adiabat Thanks for merging.

PR #29, opened 2017-03-07
Spvflip

adiabat:
Big reorg of wallet functionality.

Previously, the uspv package handled all wallet functions.  With this PR, the wallet function is now split up into two packages using interfaces:

wallit deals with storing and retreiving utxos, creating and signing transactions.

uspv deals with the network layer, sending network messages and filtering what to hand over to wallit.

qln's litNode now talks to wallit, without being aware of uspv.  The idea is that a litNode should be able to support multiple wallits now.  It's interface-y so different wallits could be used, and wallits don't need to use uspv.  An RPC fullnode connection could be used, or (ahem) a web explorer interface instead.

Seems to work; I'm sure there are bugs, but not obviously more than the previous setup, so I'll just merge it.


adiabat:
Also total code diff is like minus a kiloline so that's cool.

PR #31, opened 2017-03-09
LisAdr

jsphkn:
- Added listening address to ls display (it will only be displayed after "lis" command is run)


adiabat:
I tried it out, looks like it works well.  Also could set up for listening on multiple ports / devices which we may support later.

Cool, thanks!

PR #32, opened 2017-03-12
Collecting logging information to generate a log file

vinchang2015:
* A logfile is generated when a local machine starts listening to the server
* logging information is written to the logfile every the program calls log functions

### DIFF
--- a/lit.go
+++ b/lit.go
@@ -194,7 +194,7 @@ func main() {
 			for {
 				// setup reader with max 4K input chars
 				reader := bufio.NewReaderSize(os.Stdin, 4000)
-				fmt.Fprintf(color.Output,"LND# ")                 // prompt
+				fmt.Printf("LND# ")                 // prompt
 				msg, err := reader.ReadString('\n') // input finishes on enter key
 				if err != nil {
 					log.Fatal(err)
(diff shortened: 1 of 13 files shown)


adiabat:
Thanks Vincent-

first thing to fix is to rebase the changes onto the current master branch.  The uspv package was changed and split into uspv and wallit but this PR is based on the old master branch.  That's probably not that big of a change.

uspv/keyderiv.go
```diff
@@ -1,7 +1,7 @@
 package uspv
 
 import (
-	"fmt"
```
adiabat: Since keyderiv is interactive, these aren't suitable to be done via log.

It might make sense to remove key unwrapping from the lit UI, as it's the only interactive component left, and have unwrapping be done via RPC.  But that's another issue; for now we can leave this as interactive.

lit.go
```diff
@@ -109,8 +110,13 @@ func setConfig(lc *LitConfig) {
 }
 
 func main() {
-	fmt.Printf("lit node v0.0\n")
-	fmt.Printf("-h for list of options.\n")
+	f, err := os.OpenFile("logfile", os.O_RDWR|os.O_CREATE|os.O_APPEND, 0666)
+	defer f.Close()
+	mw := io.MultiWriter(color.Output, f)
```
adiabat: We don't want colorized text in log file output; color can stay in the interactive ui only.

vinchang2015:
> What was the rationale for this find/replace?
> Is it related to your use of log for simultaneous log file and terminal output?

Yes, and we're making the logging collection process consistent over different places.

vinchang2015:
The comments are addressed. It's easier to create a new branch #34 than rebasing the change. PTAL. Thanks.

PR #36, opened 2017-03-30
Lit bamf init commit

Firescar96:
A web gui exists, although right now it just logs all responses from lit instead of displaying them in a pretty way


lit.go
```diff
@@ -130,7 +226,16 @@ func main() {
 		log.Fatal(err)
 	}
 
-	litrpc.RpcListen(node, conf.rpcport)
+	rpcl := new(litrpc.LitRPC)
```
adiabat: Why move the RPC stuff here from litRPC?  Seems to fit better there

Firescar96: Because the mutex for shutdown depends on `rpcl`. It was weird having litrpc create `rpcl` and then provide the lock for the whole program.

litbamf/server.go
```diff
@@ -0,0 +1,21 @@
...
+	"fmt"
+	"net/http"
+	"os"
+	"path"
+)
+
+func BamfListen(port uint16) {
+
+	listenString := fmt.Sprintf("127.0.0.1:%d", port)
+	ex, err := os.Executable()
```
adiabat: replace 13 to 18 with 
```
exPath := path.Dir(os.Args[0])
```
though this still makes the js relative to executable path which won't work for most uses

Firescar96:
so:
-change os.Executable to os.Args[0]
-change to use a react cdn instead of the react packages
-split out the changes to lit help text into a separate pull request

adiabat:
Cool.  Great start to UI for the masses :)

PR #39, opened 2017-04-12
Lit Message Interface w/ implementations

thehenoch:
Changing LitMsg protocol in code
Clean up to stop having to diagnose values through splicing - instead use structs that store values
New Constructors for creating message structs that can take either values as params or a []byte and can analyze it


lnutil/msglib.go
```diff
@@ -1,47 +1,770 @@
...
+	t.Text = text
+	return t
+}
 
-	MSGID_CLOSEREQ  = 0x40 // close channel
-	MSGID_CLOSERESP = 0x41
+func NewChatMsgFromBytes(b []byte, peerid uint32) (*ChatMsg, error) {
+	c := new(ChatMsg)
+	c.PeerIdx = peerid
+	b = b[1:]
```
adiabat: *almost* no way to error, but if b is 0 (or 1) bytes long this will crash the program; put in a
```
if len(b) < 2 { return nil, fmt.Errorf("got %d bytes, expect 2 or more", len(b)) }
```
or equiv

lnutil/msglib.go
```diff
@@ -1,47 +1,770 @@
...
+	pr.HAKDbase = HAKD
+	return pr
+}
+
+func NewPointRespMsgFromBytes(b []byte, peerid uint32) (*PointRespMsg, error) {
+	pm := new(PointRespMsg)
+	b = b[1:] // get rid of messageType
+	pm.PeerIdx = peerid
+
+	if len(b) != 99 {
```
adiabat: similarly here, move the len check up to the top, before the truncation. (set to 100 I guess)

lnutil/msglib.go
```diff
@@ -1,47 +1,770 @@
...
+	cd.ElkTwo = ELKTwo
+	return cd
+}
+
+func NewChanDescMsgFromBytes(b []byte, peerid uint32) (*ChanDescMsg, error) {
+	cm := new(ChanDescMsg)
+	b = b[1:] // get rid of messageType
+	cm.PeerIdx = peerid
+
+	if len(b) != 250 {
```
adiabat: same here

lnutil/msglib.go
```diff
@@ -1,47 +1,770 @@
...
+		return nil, fmt.Errorf("got %d byte channel description, expect 250", len(b))
+	}
+
+	var op [36]byte
```
adiabat: this works but can get annoying if you need to change things.  bytes.buffer can make this easier, because you can use buf.Next(33) to read out the later bytes.  Either way works though.

adiabat:
Also, with a lot of git commits like this, I think it'll be easier to review with squashed commits.  you can keep a branch with individual commits on your side to keep track.  Here's a guide:
http://eli.thegreenplace.net/2014/02/19/squashing-github-pull-requests-into-a-single-commit

adiabat:
superceded by #42

PR #43, opened 2017-04-28
Use adiabat packages

jnewbery:
This PR removes the btcsuite/btcd and btcsuite/btcutil imports and replaces them with imports from adiabat/btcd and adiabat/btcutil.

btcsuite does not currently support segwit. Once segwit support is merged (https://github.com/btcsuite/btcd/pull/656) we can revert the imports to btcsuite so we pick up the latest version.

README also updated since it's no longer required to add new git remotes to get lit to build.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,32 +12,8 @@ Under development, not for use with real money.
 
 3. Download the lit project: `go get github.com/mit-dci/lit`
 
-4. The `go get` will fail. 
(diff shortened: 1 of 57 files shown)


adiabat:
Cool, thanks!   Hopefully will be able to switch back to upstream btcsuite once segwit is merged in (hint @roasbeef :)

Issue #44, opened 2017-04-28
lndc_test.go fails


jnewbery:
running lndc_test.go fails:

```
→ go test .
# github.com/mit-dci/lit/lndc
lndc/lndc_test.go:43: cannot use localPriv.PubKey().SerializeCompressed() (type []byte) as type string in argument to conn.Dial
```

I've tried changing the []byte to a string in call to `conn.Dial()`:

```
@@ -40,7 +40,7 @@ func TestConnectionCorrectness(t *testing.T) {
        wg.Add(1)
        go func() {
                dialErr = conn.Dial(remotePriv, listener.Addr().String(),
-                       localPriv.PubKey().SerializeCompressed())
+                       string(localPriv.PubKey().SerializeCompressed()))
                wg.Done()
        }()
```

but lndc_test now times out after 10 minutes and fails:

```
→ go test -v .
=== RUN   TestConnectionCorrectness
SIGQUIT: quit
PC=0x45af01 m=0 sigcode=0

goroutine 0 [idle]:
runtime.futex(0x95ded0, 0x0, 0x0, 0x0, 0x7f5000000000, 0xa42cd940, 0x0, 0x0, 0x7ffea42cd970, 0x40f32b, ...)
	/usr/local/go/src/runtime/sys_linux_amd64.s:422 +0x21
runtime.futexsleep(0x95ded0, 0x0, 0xffffffffffffffff)
	/usr/local/go/src/runtime/os_linux.go:45 +0x62
runtime.notesleep(0x95ded0)
	/usr/local/go/src/runtime/lock_futex.go:145 +0x6b
runtime.stopm()
	/usr/local/go/src/runtime/proc.go:1650 +0xad
runtime.findrunnable(0xc420022600, 0x0)
	/usr/local/go/src/runtime/proc.go:2102 +0x2e4
runtime.schedule()
	/usr/local/go/src/runtime/proc.go:2222 +0x14c
runtime.park_m(0xc4204309c0)
	/usr/local/go/src/runtime/proc.go:2285 +0xab
runtime.mcall(0x7ffea42cdb00)
	/usr/local/go/src/runtime/asm_amd64.s:269 +0x5b

goroutine 1 [chan receive, 9 minutes]:
testing.(*T).Run(0xc42047e270, 0x6f1db0, 0x19, 0x81f880, 0xc42043fd20)
	/usr/local/go/src/testing/testing.go:698 +0x2f4
testing.runTests.func1(0xc42047e270)
[...]

PR #45, opened 2017-04-28
Add minimal Travis support

jnewbery:
This PR adds Travis support. Currently, all Travis will do is build lit and run the unit tests. Future PRs should:

- add integration tests to be run by Travis
- add code coverage
- etc

@adiabat - to get this to run, you'll need to open an mit-dci travis account (free for open source projects), and then update the settings to do CI on the lit project. You'll also need to enable Travis on the github side from the mit-dci/lit repository settings → Integration and services → services if you want Travis to run on PRs.

Travis runs currently fail the unit tests due to #44

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,7 @@
+language: go
+go:
+  - 1.8
+sudo: false
+script:
+  - go get -v ./...
+  - go test ./...


adiabat:
Cool, tests should be working.  Will try to keep that the case :)

jnewbery:
:+1: I like tests. Especially ones that get run. And even more especially ones that get run and pass.

PR #47, opened 2017-05-04
add rpc and database code to support peer nicknames

Firescar96:


litrpc/netcmds.go
```diff
@@ -64,6 +65,25 @@ func (r *LitRPC) Connect(args ConnectArgs, reply *StatusReply) error {
...
+	Nickname string
+}
+
+func (r *LitRPC) AssignNickname(args AssignNicknameArgs, reply *StatusReply) error {
+	err := r.Node.SaveNicknameForPeerIdx(args.Nickname, args.Peer)
+	if err != nil {
+		return err
+	}
+
+	r.Node.RemoteCons[args.Peer].Nickname = args.Nickname
```
adiabat: This can crash if the peer number doesn't exist.
First do
```
peer, ok := r.Node.RemoteCons[args.Peer]
if !ok{
return fmt.Errorf("peer %d doesn't exist", args.Peer)
}
```

Firescar96: I don't think so, in that case this error would be returned by the SaveNicknameForPeerIdx call above
https://github.com/mit-dci/lit/pull/47/files#diff-879af1248dab96ee2c9e0625c12f4bfeR281
and AssignNickname would already exit

qln/lndb.go
```diff
@@ -126,6 +127,7 @@ func (inff *InFlightFund) Clear() {
 func (nd *LitNode) GetPubHostFromPeerIdx(idx uint32) ([33]byte, string) {
 	var pub [33]byte
 	var host string
+
```
jnewbery: No need to change whitespace here.

qln/magicnums.go
```diff
@@ -25,8 +25,9 @@ var (
 	BKTChanMap = []byte("cmp") // map of channel index to outpoint
 	BKTWatch   = []byte("wch") // txids & signatures for export to watchtowers
 
-	KEYIdx  = []byte("idx") // index for key derivation
-	KEYhost = []byte("hst") // hostname where peer lives
+	KEYIdx      = []byte("idx")  // index for key derivation
+	KEYhost     = []byte("hst")  // hostname where peer lives
+	KEYnickname = []byte("nick") // hostname where peer lives
```
jnewbery: Bad comment

qln/lndb.go
```diff
@@ -173,7 +209,7 @@ func (nd *LitNode) NextChannelIdx() (uint32, error) {
 	return cIdx, nil
 }
 
-// GetPeerIdx returns the peer index given a pubkey.  Creates it if it's not there
+// GetPeerIdx returdns the peer index given a pubkey.  Creates it if it's not there
```
jnewbery: typo introduced

qln/lndb.go
```diff
@@ -214,11 +250,48 @@ func (nd *LitNode) GetPeerIdx(pub *btcec.PublicKey, host string) (uint32, error)
 				return err
 			}
 		}
+
```
jnewbery: Don't change whitespace here

qln/lndb.go
```diff
@@ -214,11 +250,48 @@ func (nd *LitNode) GetPeerIdx(pub *btcec.PublicKey, host string) (uint32, error)
...
 		return nil
 	})
 	return idx, err
 }
 
+// GetPeerIdx returdns the peer index given a pubkey.  Creates it if it's not there
+// yet!  Also return a bool for new..?  not needed?
```
jnewbery: Bad comment

qln/lndb.go
```diff
@@ -154,6 +156,40 @@ func (nd *LitNode) GetPubHostFromPeerIdx(idx uint32) ([33]byte, string) {
 	return pub, host
 }
 
+// GetNicknameFromPeerIdx gets the pubkey and internet host name for a peer
```
jnewbery: Bad comment

qln/lndb.go
```diff
@@ -214,11 +250,48 @@ func (nd *LitNode) GetPeerIdx(pub *btcec.PublicKey, host string) (uint32, error)
...
 		}
+
 		return nil
 	})
 	return idx, err
 }
 
+// GetPeerIdx returdns the peer index given a pubkey.  Creates it if it's not there
+// yet!  Also return a bool for new..?  not needed?
+func (nd *LitNode) SaveNicknameForPeerIdx(nickname string, idx uint32) error {
```
jnewbery: `SetNicknameForPeerIdx()` is probably a better name.

litrpc/netcmds.go
```diff
@@ -64,6 +65,25 @@ func (r *LitRPC) Connect(args ConnectArgs, reply *StatusReply) error {
...
 }
 
+// ------------------------- name a connection
+type AssignNicknameArgs struct {
+	Peer     uint32
+	Nickname string
+}
+
+func (r *LitRPC) AssignNickname(args AssignNicknameArgs, reply *StatusReply) error {
+	err := r.Node.SaveNicknameForPeerIdx(args.Nickname, args.Peer)
```
jnewbery: Can you add a comment on why you need to call SaveNicknameForPeerIdx and then also set Nickname on r.Node.RemoteCons?

litrpc/netcmds.go
```diff
@@ -64,6 +65,29 @@ func (r *LitRPC) Connect(args ConnectArgs, reply *StatusReply) error {
...
+	if err != nil {
+		return err
+	}
+
+	peer, ok := r.Node.RemoteCons[args.Peer]
+	if !ok {
+		return fmt.Errorf("peer %d doesn't exist", args.Peer)
```
adiabat: sorry, looking at this further, this isn't actually an error.  If the first call to `SaveNicknameForPeerIdx() ` fails, that's an error.  But of the second part, modifying `RemoteCons[args.Peer]` fails, that just means you assigned a nickname for a peer you're not currently connected to, which is OK.

So maybe it's better to move the `reply.Status` above this, and then just `if !ok {return nil}`

Firescar96:
in other words, I should review my own pull requests first

qln/lndb.go
Firescar96: i saw some other functions here that used save, pero no me da igual

litrpc/netcmds.go
Firescar96: You might be on to something though, I was able to crash my node by sending a request a call to AssignNickname, I didn't note where the error was, I'll try again.

qln/magicnums.go
Firescar96: I changed my editor settings so now I can read comments :sweat_smile:

litrpc/netcmds.go
Firescar96: I see what you're saying, it's fixed in the upcoming commit

Firescar96:
bump

Firescar96:
it looks like this was integrated into master already

PR #48, opened 2017-05-04
Lit bamf channels page

Firescar96:
update to channels page


adiabat:
This only modifies litbamf js code so utACK

PR #49, opened 2017-05-04
Cleanup litpy

jnewbery:
This PR refactors the litpy functions into a LitConnection class.

Several other cleanups, including adding a README and fixing the PEP8 warnings from flake8.


adiabat:
Looks great, I am not super well-versed in python so help here is very appreciated!

PR #50, opened 2017-05-04
Add integration test and run in travis

jnewbery:
Very basic for now - starts a lit node, creates an address, checks the balance and shuts down the node.

This at least builds and runs a lit node in CI. That should catch any egregious errors that cause lit to fail to start.

More thorough testing can be added later.


adiabat:
Looks good

PR #46, opened 2017-05-03
Basic lightning integration test

vinchang2015:
- creating a regtest node to support mineblock(), sendTo(), getInfo() functional calls
- creating a litconn node to support litNewAddress(), litSend(), getWitAddress(), getLegacyAddress() and getBal() functional calls
- creating the first test_receive file to test the process of setting up a regtest node and a lit node and the ability of a lit node to receive money from the regtest node

PR #54, opened 2017-05-06
[integration test] A litnode receiving Bitcoin from regtest network

vinchang2015:
- Built upon #46, combined with the latest litrpc.py and test_lit.py updates
- Updated the testkey.hex file to use random numbers

### DIFF
--- a/cmd/litpy/test_lit.py
+++ b/cmd/litpy/test_lit.py
@@ -3,6 +3,8 @@
 import os
 import subprocess
 import tempfile
+import random
+from string import digits
 
 import litrpc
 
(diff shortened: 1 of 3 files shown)


vinchang2015:
@jnewbery: this is the version rebased with your latest changes 
PTAL when you have a time

cmd/litpy/test_lit.py
```diff
@@ -17,7 +19,8 @@ def __init__(self, i):
 
         # Write a hexkey to the hexkey file
         with open(self.data_dir + "/testkey.hex", 'w+') as f:
-            f.write("1" * 64 + "\n")
+            randomString = ''.join(random.choice("abcdef"+digits) for i in range(64))
```
adiabat: It makes sense for the key to be random though; if creating multiple nodes with the same testkey.hex, they will all have the same addresses which can lead to lots of problems.

jnewbery: Ah good point. Makes sense that this needs to be different for each node.

In my opinion, we'd still want this to be deterministic instead of random though - perhaps the hex value of the node id. Generally determinism is useful in functional tests for repeatability.

jnewbery:
@vinchang2015 I've incorporated your changes to the test script into #57 so I think we can close this PR.

vinchang2015:
Thanks @jnewbery

PR #56, opened 2017-05-08
enable synchronous multiclient support for chat messages

Firescar96:


qln/lndb.go
```diff
@@ -88,8 +88,10 @@ type LitNode struct {
 	// Nodes don't have Params; their SubWallets do
 	// Param *chaincfg.Params // network parameters (testnet3, segnet, etc)
 
-	// queue for async messages to RPC user
-	UserMessageBox chan string
+	// websocket connections of users to a message channel
+	UserWsCons map[int]chan lnutil.ChatMsg
```
adiabat: (granted, UserMessageBox channel isn't a good way to do it & needs to change, BUT...)

UserWsCons in qln isn't the right way to structure this.  All the functionality in the qln package should have no idea about RPC, be it websockets or otherwise.  The litrpc package is the right place to put things that deal with websockets.  It's quite possible we'll want to make a standalone binary (maybe for phones?) which doesn't use RPC at all and has everything integrated into 1 binary.  In such a version, litrpc wouldn't exist at all, but qln should work unmodified.

qln/msghandler.go
```diff
@@ -53,6 +47,28 @@ func (nd *LitNode) PeerHandler(msg lnutil.LitMsg, q *Qchan, peer *RemotePeer) er
...
+// BroadcastMessage sends a chat message to all connected websocket clients
+func (nd *LitNode) BroadcastChatMessage(msg lnutil.LitMsg) error {
+	chat, ok := msg.(lnutil.ChatMsg)
+	if !ok {
+		return fmt.Errorf("can't cast to chat message")
+	}
+
+	// get the current set of channels and save them, they may be deleted by litrpc
+	// before the iteration is over
+	var channels []chan lnutil.ChatMsg
```
adiabat: avoid making a variable named "channels"

unfortunately there's already a collision in the go language and lightning channels; I call all lightning channels "qchannels" because it doesn't collide with the `chan` in go, and nothing else starts with q.

Also... why have 2 range loops? just say 
`c <- chat`
in the first one...

qln/lndb.go
Firescar96: I could change the name to `UserChatChans` because like UserMessageBox it doesn't actually contain any websockets. It has to be in qln instead of litrpc so that qln/msghandler.go has access to it.

qln/msghandler.go
Firescar96: Because of this: https://github.com/mit-dci/lit/pull/56/files#diff-4cec2aab01ce8eee14372ee15be1af85R100
I'm thinking about a possible race condition where if I only use one loop the function will never end because if there are multiple users they may delete their old chan and make a new one so quickly that the loop never ends.
I tried to reference this in the comment.

Firescar96: or would you recommend making this explicit with a mutex

adiabat: wouldn't two loops make the problem worse? Instead of ranging over the map of UserWsCons and sending a message to each, this ranges over UserWsCons, caches each one, then starts over and ranges over them to send messages.  It seems like there's more time for things to close during the 'seek back to the start' part.

Firescar96: it shouldn't, the concern is with the `range nd.UserWsCons` where nd.UserWsCons changes during the iteration
`range channels` doesn't have the same problem

Firescar96: I will mention that at a point while I was trying to write this code it did contain websockets and that's where the name comes from

adiabat: golang channels can also close; if the websocket connection closes, the golang channel does as well, right?
what if nd.UserWsCons changes *after* iterating over it?  That seems more likely than if it changes during iteration.  And then you've got a reference to a closed go channels that you try to push messages to.
https://dave.cheney.net/2014/03/19/channel-axioms
sends to a closed channel panic / crash the program.

Firescar96: The chan is never closed, after it's used I'm expecting golang garbage collection to take care of it But each chan is only ever used once, when a peer wants a new message they make a new chan.

and it's expected for nd.UserWsCons to change after iterating over it, it's supposed to change as in litrpc clients are removing a reference to the old channel and creating a reference to a new channel in nd.UserWsCons.

qln/init.go
```diff
@@ -4,10 +4,10 @@ import (
 	"fmt"
 	"path/filepath"
 
-	"github.com/boltdb/bolt"
 	"github.com/adiabat/btcd/chaincfg"
 	"github.com/adiabat/btcd/wire"
 	"github.com/adiabat/btcutil/hdkeychain"
+	"github.com/boltdb/bolt"
```
Firescar96: my editor does this automatically

cmd/lit-af/wsRPC.go
```diff
@@ -0,0 +1,115 @@
+package main
```
Firescar96: a large part of this was taken from https://golang.org/src/net/rpc/jsonrpc/client.go but I'm not sure how to properly cite it

Firescar96:
the way this work should be more intuitive:
when a the lit binary receives a chat message msghandler.go adds it to a queue
then it gets picked up buy a goroutine in litrpc and forwarded to all connected users

maybe someone else has a method of doing this which doesn't require rewriting some codec code?

Firescar96:
@adiabat I updated the PR so it doesn't break json rpc clients and uses a polling based method of chat retrieval

metalicjames:
add to whitelist

Firescar96:
oh yeah, is this applicable anymore? I can work on it again

metalicjames:
What is it supposed to do?

Firescar96:
At the time of development lit couldn't handle multiple clients being connected.eg. Lit-BAMF open in two browser windows at once. So this PR changed messaging so multiple clients could subscribe to messages and each get a copy.

An alternative implementation would simply not support multiple clients.

metalicjames:
Given that Lit does now seem to support multiple simultaneously connected RPC clients, is this PR now redundant?

Firescar96:
Yup, good work Lit devs.

PR #57, opened 2017-05-08
test_lit.py improvements

jnewbery:
General improvements to test_lit:

- I've incorporated @vinchang2015's new receive test case.
- added litecoin support in anticipation of multichain wallit support
- moved the integration test into its own base directory

notes:

- travis doesn't like me at the moment (I assume because this has triggered a bitcoin mining flag). I've emailed them to re-enable my account. Don't be surprised if Travis refuses to build this PR.
- this seems a little flakey - sometimes the receive funds fails. I haven't dug into why that is

This should replace #54 since it contains a superset of the functionality in that PR.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,5 +16,5 @@ script:
   - go get -v ./...
   - go test ./...
   - go build
-  - cmd/litpy/test_lit.py
+  - test/test_lit.py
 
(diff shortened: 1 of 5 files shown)


adiabat:
Looks good -- hopefully we can get Travis to accept the bitcoind requirements.  If not, we can set up a different testing framework.

PR #59, opened 2017-05-08
Multi-wallet

adiabat:
Tests don't pass as some RPC structures have changed.  Will describe the PR and update tests shortly.

Also this is a huge PR and hopefully there won't be more of these; will try to keep mostly in sync with master and perform smaller PRs.


adiabat:
Travis tests are gone for now but manually running test_lit.py mostly works on the lit side, but not the bitcoind side.

the test_lit.py script fails because of control over the bitcoind node.  The errors I get are:
```
Received response from bitcoin node: {"result":null,"error":{"code":-32600,"message":"Params must be an array"},"id":"2645"}
```
this may be due to a different version of bitcoind, or copying code from the lit and bitcoind rpcs which may work somewhat differently.  If it's makes sense, we can change the lit rpc calls to work more like bitcoind, which is more widely known and used.

I think I will merge this change in even with this part of the automated tests not working. This is a fairly big / disruptive change so there will probably be bugs it generates.  But I'd rather get it in to master sooner so that everyone else can base their work on it.  Will merge this in tomorrow if there aren't any objections.

jnewbery:
`"Params must be an array"` tells me your version of bitcoind predates this: https://github.com/bitcoin/bitcoin/pull/8811 , so is pre v0.14.

I'll add a check to the tests to make sure that bitcoin is at version 0.14 or above.

adiabat:
Test seems to be failing due to travis / bitcoind errors; it passes on my local machine with bitcoind 0.14.1.  Will merge this now so things can be based off it.

PR #60, opened 2017-05-09
Fix and improve tests

jnewbery:
A bunch of improvements:

- refactor BCNode, LCNode and LitNode into their own modules
- Add copyright notice and use `#!/usr/bin/env python3` hashbang instead of `#!/usr/bin/python3` for portability
- check that bitcoind version is >0.14 and litecoind version is >0.13.2
- fix build/test in travis
- cache bitcoind and litecoind in travis to avoid slow downloads
- improve logging on failure
- make sure bitcoind and lit processes are cleaned up at the end of the test

Some of these commits are broken until the `[tests] fixups` commit, but I'm not going to bother rearranging them since I expect you'll merge this as a single commit.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,13 +5,18 @@ go:
 sudo: required
 install:
   - sudo apt-get update
-  - sudo apt-get install python3 python3-pip python3-requests
-  - sudo apt-get install bitcoind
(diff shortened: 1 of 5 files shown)


adiabat:
Cool, will base other tests on this.

PR #63, opened 2017-05-10
Test framework

jnewbery:
This moves the test framework into its own lit_test_framework module, to be imported and subclassed by indvidual test scripts.

There is currently only one test script: test_basic.py. Additional tests named `test_[description].py` can be added.

This splits the test logic from the test framework so additional tests can be added and worked on in parallel.

README updated to describe the new structure.


adiabat:
looks good!

PR #64, opened 2017-05-10
Improved logging

jnewbery:
- changes lit to log microseconds
- changes bitcoind config in test framework to log microseconds
- adds combine_logs tool for combining logs from multiple lit nodes and bitcoind nodes. Output looks like:

screenshot from 2017-05-10 17-58-44

(once logging has been added to the test framework, those logs will also be included in the combined log)


adiabat:
cool -- we need microseconds to time micropayments

PR #65, opened 2017-05-10
[tests] Add channel tests

jnewbery:
Updates the basic test to:

- open a channel between two nodes
- push funds one way
- push funds back
- close the channel


adiabat:
This is where the real testing starts... and the real bugs are discovered :)

jnewbery:
I'm only here for the :bug:s

PR #66, opened 2017-05-11
Move coin specific parameters like difficulty and proof of work to btcd

metalicjames:
This pull request is a WIP. It also requires changes to btcd which are in a separate PR.

Even better would be to make chaincfg.params serializable somehow so you could load the definitions from a file. That way, adding new coins would be as easy as making a new .conf file.


adiabat:
Looks like a better way to do it; I will apply the changes to the btcd library first so that it passes tests.

Also, can you base off / merge into master instead of multiparam

coinparam/register.go
```diff
@@ -31,6 +32,26 @@ type Params struct {
...
 	GenesisHash *chainhash.Hash
+	
+	// The function used to calculate the proof of work value for a block
+	PoWFunction func(b []byte) chainhash.Hash
+  
+    // The function used to calculate the difficulty of a given block
+    DiffCalcFunction func(r io.ReadSeeker, height, startheight int32, p *Params) (uint32, error)
+  
+    // The block header to start downloading blocks from
+    StartHeader string
```
adiabat: Can you change this to [80]byte or []byte?  String is easier but could result in runtime errors rather than compile time errors.  Or why not use wire.BlockHeader?  That seems safest

coinparam/register.go
```diff
@@ -31,6 +32,26 @@ type Params struct {
 
 	// GenesisHash is the starting block hash.
 	GenesisHash *chainhash.Hash
+	
+	// The function used to calculate the proof of work value for a block
+	PoWFunction func(b []byte) chainhash.Hash
+  
+    // The function used to calculate the difficulty of a given block
+    DiffCalcFunction func(r io.ReadSeeker, height, startheight int32, p *Params) (uint32, error)
```
adiabat: the height arguments here but can leave them in for now and remove later if there are no coins using it.

metalicjames: Height here is just to inform the difficulty function which block it needs to calculate the difficulty of. The other alternative is to pass the header directly into the function, but this would not work for LTC or VTC which have a variable retarget lookback.

adiabat: Yeah, this will work, I've just got a feeling there's a better way to do it.
But I can't actually think of one so this is fine :)

coinparam/register.go
metalicjames: This is fixed now.

uspv/header.go
```diff
@@ -50,14 +37,11 @@ func checkProofOfWork(header wire.BlockHeader, p *coinparam.Params) bool {
 	}
 
 	// The header hash must be less than the claimed target in the header.
-	var blockHash chainhash.Hash
+  
+    var buf bytes.Buffer
+	_ = wire.WriteBlockHeader(&buf, 0, &header)
```
adiabat: I get an error here; did you also change adiabat/btcd? could you add the PR there and I'll merge that first?

adiabat:
Test says failed due to timeout, but it works on my machine.  Maybe we need to change the expected times for tests.

fixes lots; cool

PR #71, opened 2017-05-16
Add SetFee RPC command

adiabat:
This isn't a good long term solution, but for now it helps to be able to manually set fee rates.

There's a new RPC command SetFee which lets you set a fee rate (satoshis per byte) for any wallet.  If you give it a negative number, it won't change the fee rate.  It always returns the current fee rate.

(So if you want to just check what the fee rate is, send -1)

The corresponding command has been added to lit-af.

### DIFF
--- a/cmd/lit-af/shell.go
+++ b/cmd/lit-af/shell.go
@@ -158,6 +158,13 @@ func (lc *litAfClient) Shellparse(cmdslice []string) error {
 		}
 		return nil
 	}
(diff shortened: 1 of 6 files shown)


adiabat:
@jnewbery expressed a valid concern that this is pretty ugly.  If it offends others we can add another RPC call, "GetFee" instead of intentionally sending nonsense values to get the current fee.  For now though this will work.

PR #72, opened 2017-05-16
[wip] Test litecoin

jnewbery:
Adds a test for litecoin. At the moment, this is almost a duplicate of the test_basic test. In the future it'd be nice to refactor into a single test. An argument would be passed into the test to specify which chain should be used (or a list of chains in the case where we're testing multiple chains).

There are a few odds and ends in here as well - attaching a debugger, testing whether any of the required binaries are already running when the test starts, and cleaning up the temp directory after a test run.

Still a work in progress. Travis is failing to build this correctly. I can't tell why.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,4 +24,5 @@ script:
   - go test ./...
   - go build
   - test/test_basic.py
+  - test/test_lc.py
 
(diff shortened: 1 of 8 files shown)


jnewbery:
Hmmm ok, this passes travis on mit-dci/lit, but not on jnewbery/lit. I wonder why that is.

Whatever, this should be good to merge.

adiabat:
Cool; not sure about the different branches problem.  With SetFee we could try to get rid of the coin type based fee settings... or add a list of default fees to start with on different coin types.
For testing though it could make sense to add SetFee calls to the tests.

jnewbery:
> this passes travis on mit-dci/lit, but not on jnewbery/lit. I wonder why that is.

I think this is fixed by https://github.com/mit-dci/lit/pull/75

PR #73, opened 2017-05-16
[rpc] add Fee() and SetFee() rpcs

jnewbery:
Adds explicit getter and setter functions for fee. Follows #71


adiabat:
Ohhh ok.  I guess we should have things be clean and sensible.
(extra 50 lines of code though... at what cost readability? :))

PR #77, opened 2017-05-17
[tests] break channel

jnewbery:
Add a new test that breaks a channel:

- setup, create channel, send funds around
- node0 breaks channel
- assert node0 has funds but can't spend, node1 has funds and can spend
- advance 5 blocks
- assert node0 and node1 can both spend funds


adiabat:
great.  This is another one of those tests I'm am less than 100% sure of, in that I've seen this functionality not work in some cases.  Great to have a deterministic test of it here.

PR #78, opened 2017-05-18
[rpcs] Better ChanTotal in Balance()

jnewbery:
ChanTotal currently includes the channel balance for channels that are closed. Don't do that.


adiabat:
This certainly makes sense.  It's a wonder this problem existed in the first place... :)

PR #88, opened 2017-09-15
Basic Config File Parsing

Varunram:
This PR aims to accomplish config file parsing to some extent. As of now, it can parse params in `config/`. Will keep constantly updating this branch under @adiabat's suggestions and inputs.

### DIFF
--- a/lit.conf
+++ b/lit.conf
@@ -1,0 +1,3 @@
+; Use this as a comment. Specify all parameters below similar to those what you would on the CLI
+rpcport=8012
+reg=localhost
--- a/lit.go
+++ b/lit.go
@@ -1,7 +1,9 @@
 package main
 
 import (
-	"flag"
+	"bufio"
+	"crypto/rand"
+	"encoding/base64"
 	"fmt"
 	"io"
 	"log"
@@ -10,6 +12,7 @@ import (
 	"strings"
 	"time"
 
+	flags "github.com/jessevdk/go-flags"
 	"github.com/mit-dci/lit/coinparam"
 	"github.com/mit-dci/lit/litbamf"
 	"github.com/mit-dci/lit/litrpc"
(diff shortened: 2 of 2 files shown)


adiabat:
You put your privatekey in here, also a binary file.  Might want to get rid of that :)

I'll look at viper more, but I think the code from btcd handling the config file looks like a better fit for this use case.

Varunram:
Thanks a ton @adiabat and sorry!

Varunram:
Yeah, makes sense. I'll checkout btcd, prune their stuff and update this PR :)

Varunram:
@adiabat Updated updated!

adiabat:
Cool, thanks for writing this! 

One feature I use quite a lot is the -dir flag, to specify a directory for all the lit files.  For normal usage it's probably not that important, but right now it's useful for testing by having multiple lit nodes running on a single machine.  Right now this config file PR doesn't work nicely with that feature; it seems to only work with the default datadir, $HOME/.lit.  

I think the most straightforward way to support custom directories would be to parse command line args first, and if it sees a -dir flag, to look for a lit.conf in that directory along with everything else.

If that seems annoying we can merge this first with only single directory support; in fact maybe for the custom directory case, having a config file isn't as important.  However, the -dir flag should still work, and if it can't find the lit.conf, it ignore it and not require one.

Varunram:
Done, a couple of small things though, the `-dir` flag must be specified as a whole instead of just being relative to the binary (we could change that, but this would address more use cases). Also, I decided to remove the `-c` option earlier as I thought it didn't make sense to have the config file and  home directories in separate places. If the config file has been deleted in the specified home directory, a new default config file will be generated.

Varunram:
Rebased on master. I did my best to incorporate the tower code in here, please do tell if there are any unintended bugs over here. Thanks!

adiabat:
Looks good!  I will add / modify this a bit afterwords but this looks good.  I'll probably add a hardcoded starting settings lit.conf with comments in it for people to edit.

PR #92, opened 2017-09-28
Give preference to CLI params

Varunram:
Deep Apologies for missing this as I was testing. Rearranged the config parsing stuff and removed one redundant variable and piece of code.


adiabat:
ok... will try to figure out test framework later

PR #93, opened 2017-09-30
Connect to DNS Seeds if user specifies 1/auto/true

Varunram:
Kinda basic, but putting out here so that I know that this behaviour is intended. Regarding pinging a remote node for info, I just added a single line one `eight333.go` for now for testing but it looks like I didn't get the switch right with `wire.MsgGetAddr` in `msghandler.go`. Since there are no docs attached for this, all I did was trial and error. Right now, the DNSSeeding stuff is located over at `lit.go` and not `init` in uspv (since not all the chains actually have elements in the dns seeds). I'll move it to a better place to avoid the repeating lines once everything's cool. @adiabat

### DIFF
--- a/lit.go
+++ b/lit.go
@@ -85,6 +85,7 @@ func linkWallets(node *qln.LitNode, key *[32]byte, conf *config) error {
 			return err
 		}
 	}
+	// try testnet3
 	if conf.Tn3host != "" {
 		p := &coinparam.TestNet3Params
 		if !strings.Contains(conf.Tn3host, ":") {
@@ -108,7 +109,7 @@ func linkWallets(node *qln.LitNode, key *[32]byte, conf *config) error {
 			return err
 		}
 	}
-	// try litecoin tn4
+	// try litecoin testnet4
 	if conf.Lt4host != "" {
 		p := &coinparam.LiteCoinTestNet4Params
 		if !strings.Contains(conf.Lt4host, ":") {
@@ -134,7 +135,6 @@ func linkWallets(node *qln.LitNode, key *[32]byte, conf *config) error {
 			return err
 		}
 	}
-
 	// try vertcoin mainnet
 	if conf.Vtchost != "" {
 		p := &coinparam.VertcoinParams
--- a/lit.go
+++ b/lit.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"io"
 	"log"
+	"net"
 	"os"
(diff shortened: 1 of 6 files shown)


adiabat:
Thanks for the PR @Varunram 
Having duplicate code in lit.go before setup does work, but it can get repetitive for any other networks people want to add and connect to.  I think it would make more sense to have the code in lit.go moved into `(s *SPVCon) Connect`, where I think it could work fine.  Maybe add some kind of error if there is no `DNSSeeds[]` in the params.

`wire.MsgGetAddr` I haven't used but I get the idea of how it works.  If we receive a `wire.MsgGetAddr` (line 27 of msghandler.go in this PR) we don't need to respond at all; lit is not a full node so nobody should be asking us for peers anyway.  It's handling `wire.MsgAddr` that is more important (line 30); right now we just report the contents of the message and then drop it.  So that's where to look at for start tracking & saving remote nodes.

Varunram:
Done! One thing I'm not too sure is where to store the remote nodes the client is connected to. Once we figure that out, the road is clear :)

Varunram:
Updated!

adiabat:
Cool, some comments about the new update:

There's an empty seeds.json file in the git commit, which probably shouldn't be there as the seeds are already in coinparam.

Running the code, lit requires a file in litdir/dnsseeds/seeds.json, and will crash if the file isn't present.  Instead, it should open an empty file if nothing is there.

Since there's only 1 file, and I don't foresee there being other peer related files, there probably isn't a need for a separate directory.  Instead, the file can be stored in the subdir for each coinparam.  This makes sense because there won't be one global list of peers; the IP peers for testnet3 are separate from mainnet peers and can be stored there alongside header.bin and utxo.db.

Lastly, a bit of a nitpick but the file shouldn't be called "seeds", as the DNS seeds are hard-coded in the coinparams files.  Instead it could be "peers.json" or something like that.

When running it though, seems cool and the json formatting works.

Code-wise, `incomingMessageHandler()` is basically a big switch statement based on the message type coming in.  The case statements are 2 or 3 lines long because they hand the message off to an handler function (`TxHandler()`, `InvHandler()`, etc).  That would make sense to do here as well, with the code from line 34 down being moved into `AddrListHandler()` or a similar function.

Thanks and hope this makes sense!  I'm online if you have other questions or want clarification.

Varunram:
Closed in favour of #108 :)

PR #94, opened 2017-10-10
Small typo over at qln/lndb.go

Varunram:
Correct chanel to channel. I could combine this with the Seeds PR too, your call :)


adiabat:
ah, thanks.  Easy to merge this in on its own.

PR #96, opened 2017-10-21
Implement LN node tracker for IP-less connection

metalicjames:
Uses this simple node.js script (https://github.com/metalicjames/lit-tracker) to announce its LN address and IP signed by its public key. The server can be configured using the `tracker` option in the config file.


adiabat:
Neato!  Tried it out, seems like it works!  Probably should add some of this to the integration tests at some point.  Also a go tracker could be nice so maybe I'll do that at some point, because node, ew :)

Thanks!

PR #97, opened 2017-10-21
Switch to chain containing more work

Varunram:
Switch to the chain containing more work from the longer chain mechanism that we follow currently.

### DIFF
--- a/uspv/header.go
+++ b/uspv/header.go
@@ -276,8 +276,9 @@ func CheckHeaderChain(
 		// TODO check for more work here instead of length.  This is wrong...
 		// if we've been given insufficient headers, don't reorg, but
 		// ask for more headers.
-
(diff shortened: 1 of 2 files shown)


adiabat:
Added some comments; testing with regtest and invalidateblock, I can still get reorgs to work based on height, but I get the `Chains don't start with the same header. Quitting.!` message every time, so it's seems like the PoW comparison isn't working.

Varunram:
Hm, that seems strange.. That should fire in the case both chains don't start off with the same merkle root. Did the chains you were testing with have the same root?

adiabat:
Yeah I was just doing re-orgs on regtest using `invalidateblock`, then `generate`, got that message every time.  Do you get different behavior on yours?

metalicjames:
retest this please

PR #99, opened 2017-10-24
Update reamde and walkthrough to reflect recent changes

Varunram:


adiabat:
Ah! Thanks, this is great; gotta keep the documentation synced up with the binaries.

PR #101, opened 2017-10-30
Ignore generated .exe from windows builds

sudheesh001:


sudheesh001:
@adiabat Could you take a look into this and merge it? 😄 The build process generates `.exe` files on windows and shows them as unstaged files to be committed. These files instead shouldn't be committed.

adiabat:
Good catch, thanks.  Not many windows users around here so it's not well tested on windows, so thanks for testing in with that OS!

Issue #100, opened 2017-10-30
Usage of -tip not provided but used to explain reSync


sudheesh001:

Varunram:
Thanks for pointing out. `-tip` used to exist in earlier versions and reSync used to go along with that. I'll update the README reflecting the changes.

Varunram:
An old issue isn't always solved :)  PR still up for this, will rebase that one.

PR #102, opened 2017-10-30
Using tables for better readability

sudheesh001:
https://github.com/sudheesh001/lit/blob/bdfdbd898252e27c2831813f8568e52ed17e8843/README.md

PR #104, opened 2017-10-30
Remove --reSync

Varunram:
w.r.t #100 , since `-tip` is deprecated for now, let's remove reSync and add it back once we add `-tip`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -97,11 +97,6 @@ use <folderPath> as the directory.  By default, saves to ~/.lit/
 
 listen for RPC clients on port <portNumber>.  Defaults to 8001.  Useful when you want to run multiple lit nodes on the same computer (also need the -dir option)
 
--r or --reSync
-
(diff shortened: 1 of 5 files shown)


sudheesh001:
This might conflict with the updates in #102 which updates the readme documentation to become more readable.  

Also just a thought, I know I am not a committer to say this, why could pull requests in general not be done in one commit? makes the codebase history feel much cleaner :smile:

Varunram:
I think the idea behind commits is to reflect major changes done to the codebase. If the PR has many commits and deserves to be squashed, I think it will be. But in the above case, a single PR is reflecting changes to two issues, hence the two commits :) Regarding #102, I did see the PR, but I'm *personally* against tables (since people not only view readmes on github but on the cli too). Anyways, in case #102 is merged, it should be a single line change, that's not an issue haha

Varunram:
Closing this as we decided to add `--tip` back.

Issue #103, opened 2017-10-30
Sync blocks from tip height


Varunram:
`--reSync` is a remnant of the old `-tip` flag that used to exist before 7e1463a73826236c76b8fda58d092adc479f1a72. We should bring back `tip` in case someone finds a use for it.

PR #105, opened 2017-11-02
Add --tip to sync from given height

Varunram:
--resync now syncs the chain from the given --tip height. Other changes:

1. Change --reSync to --resync
2. Remove short argument for hard argument
3. Rename --rpcport to --port


Varunram:
Ref: #103

Varunram:
Rebased on current master.

Varunram:
Course this is relevant, will rebase after I take a look at a few other things.

Varunram:
retest this please

gertjaap:
Hey, I tested this and looked at the code.
First, my problem with the command in general is that it's a command to resync the blockchain while we have multiple wallets connected. So we're resyncing all? That seems excessive if you have a problem with just one

Secondly, the `--tip` argument is a nice touch, though if we resync all of it, then the problem is that different chains are at different heights and we only have one tip argument.

Thirdly, the tip argument is not applied to all wallits equally. So it is used in Litecoin/Vertcoin testnet/regtest/livenet but not in Bitcoin regtest/testnet.

I think we need a way to resync an individual wallit or at the very least have default tips for each and allow setting the tip for each individual one in stead of one global setting. In this setup, if I want to resync Vertcoin testnet from a particular height, it will also resync mainnet from the same height, even though testnet is at block 119k and mainnet on 920k. So resyncing testnet from 118k will do a resync of ~800k blocks on mainnet for no reason.

Varunram:
Hey, regarding the `--tip` argument, it only works against what arg we provide in the cli, e.g. `--reg` re-syncs only regtest. Of course, if you do have multiple wallets connected at the same time, this would affect you because of the `if !lnutil.NopeString(conf.Lt4host) {` check there and it would return true for multiple wallets if connected.

gertjaap:
Yeah, so the problem I see is that most people will use lit with a `lit.conf` and not supply all the chains to the commandline. So even if you only provide `--tip --resync --reg`, it will most likely also pick up the settings from the conf and resync all chains in there from the same tip.

A fix for this might be to make resync not a bool argument but a string, so that you can do `--resync=reg` or `--resync=lt4` and since it would only resync that one chain, `tip` will no longer be ambiguous

Varunram:
Will make a new PR with the suggested changes - This is too small that rebasing and building upon it might not be a wise move.

PR #106, opened 2017-11-02
Remove unused locks

### LINKED ISSUE #95: Unused Locks
While looking at #7, I came across this neat piece of TODO:
```
	nd.RemoteMtx.Lock()
	nd.RemoteMtx.Unlock() //TODO: This unlock is in the wrong place...?
```
I think we can remove this since there is nothing going on?

Varunram:
Closes #95


adiabat:
yeah, I think there was some point to that but don't remember...

PR #107, opened 2017-11-09
Add peers.json to collect a list of remote peers

Varunram:
Part [2/2] of PR #93. Creates a peers.json file in the appropriate coin sub directory (at $LITHOME) and tries to connect to the remote nodes present in it.

PR #108, opened 2017-11-09
add DNS looksup and attempts

adiabat:
implement DNS lookups, have uspv deal with appending ports


adiabat:
Some code taken from @Varunram 's #93 

Looks like it works OK, can extend with #107

PR #109, opened 2017-11-10
Cleanup lit.go

Varunram:
1. --help no longer prints twice
Remove fmt. statements and use log. wherever appropriate
Incorrect args now show "Use ./lit -h to show usage"

2. Creating a config file at new home directories does not require a sample lit.conf

### DIFF
--- a/lit.go
+++ b/lit.go
@@ -2,14 +2,11 @@ package main
 
 import (
 	"bufio"
-	"crypto/rand"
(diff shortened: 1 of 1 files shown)


adiabat:
Looks good; helps with defaults so that now just 
```$ ./lit```
will work for most people trying it out

Varunram:
Thanks!

PR #110, opened 2017-11-15
UI fixes for andui

Varunram:


adiabat:
I'd rather not need a "copy" button if users can copy with their standard UI tools, like right-clicking or ctrl-c.  If an external library is needed, then that's OK if it can't be avoided.  However in this case, it crashes for me:
```panic: No clipboard utilities available. Please install xsel or xclip.

goroutine 8 [running, locked to thread]:
main.main.func1.2(0xc4204a2820)
	/home/user/go/src/github.com/mit-dci/lit/cmd/andui/main.go:98 +0x66
github.com/andlabs/ui.doButtonOnClicked(0x1960490, 0x0)
	/home/user/go/src/github.com/andlabs/ui/button.go:112 +0x67
```

I've had no problem with copy & paste before, so this is a regression for me.  It seems like it *knows* it'll fail, so maybe just don't show the button if it knows it'll crash when you click it.

Varunram:
Oh, alright. Can't seem to reproduce, but if its introducing regressions, we might as well remove it. I can't seem to use `cmd+c` on macOS but right clicking and copying seems to work, so no worries :)

adiabat:
OK new update looks good!

PR #112, opened 2017-11-18
[WIP] Replace By Fee

Varunram:
A WIP PR to test Replace by Fee.


adiabat:
I know it's WIP / starting out, but some thoughts:

probably don't need a bool for txs or for the wallet, and can just always have RBF on.  The whole idea of "opt in RBF" is kindof silly / political thing, because you can't *prevent* RBF, even with soft forks or whatever; that's the whole point of the mining after all.  So it's kindof fooling people to pretend that you can turn off RBF.

Another thing is that the lightning commitment txs all have non-final sequence numbers.  Also the non-justice txs spending the commitment outputs also have non-final sequence because they use OP_CSV.

So probably can just leave out settings and set normal txs from the wallet to have non-final sequence.  That part's easy; then some kind of bump-fee management would be most of the code.

Varunram:
Just updated to signal RBF. Will update you more on this today :)

Varunram:
@adiabat Stuff kinda works, will catch up on this later!

wallit/signsend.go
```diff
@@ -21,10 +21,201 @@ import (
...
+				txin.SignatureScript = sigStash[i]
+			}
+			if witStash[i] != nil {
+				txin.Witness = witStash[i]
+				txin.SignatureScript = nil
+			}
+		}
+
+		log.Printf("tx: %s", TxToString(tx))
+		w.NewOutgoingTx(tx)
```
adiabat: This won't work; it will just try to broadcast txs that have locktimes in the future, which will be rejected.

PR #113, opened 2017-11-21
Dockerized LIT for easy portability

gertjaap:
I built a Dockerfile to be able to run LIT as a docker container. Not sure if this is something you guys see use for, but it makes it quite easy to run the LIT daemon on server/embedded Linux environments that are able to use docker.

What it currently does is it runs go get on the original repo to get the prerequisites and then copies in the source code files from the working directory. You could remove lines 8 and 9 to simply have it build a container out of the active master branch.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,15 @@
+FROM golang:alpine
+
+RUN apk update && apk upgrade && \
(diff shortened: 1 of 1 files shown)


adiabat:
Sounds good; I don't use docker but if other people do, makes sense to support it.  Thanks!

PR #114, opened 2017-11-22
Remove lit.conf since we auto generate conf file

Varunram:
Since we now handle conf stuff in lit.go, doesn't make sense to have this hanging around.


adiabat:
Makes sense, looks good

PR #116, opened 2017-11-23
Set dirty when the highest blockhash changes

gertjaap:
This will monitor the hash of the highest block and set the dirty flag when it changes


adiabat:
Cool, thanks!  There are other parts of the powless2 branch needed, and I will try to write those in soon and merge it to master.

PR #121, opened 2017-12-01
Correct testnet diff prediction

Varunram:
Do difficulty prediction in a specific branch the way btcd does it.


adiabat:
cool; I still think there is a remaining issue with reorgs and difficulty calculation, but this doesn't hurt and may be the fix, I'll keep running on testnet3.

PR #122, opened 2017-12-04
Better Coin Selection based on Electrum

Varunram:
Well, the idea is to implement something like the privacy approach follows in Electrum, so that we can choose coins in a better way.

As of now, the PR has a couple of helper functions and uses the current coin selection algo as fall-back (the seed is a constant right now). Will continue to update this PR.

### DIFF
--- a/wallit/helpers.go
+++ b/wallit/helpers.go
@@ -1,9 +1,23 @@
+// There is some inheritance from the base "coin" class to derived types in electrum.
+// Investigating that.. and they use buckets, not slices argh.
+// Structure of how stuff is done right now in Electrum:
+// Coin Chooser Base (base class): bucketize_coins, change_amounts, change_outputs, make_tx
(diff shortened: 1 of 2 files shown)


Varunram:
Revisiting this, might be worth to see how Core implements Murch's algo, now that it is merged over there.

alindr:
it's a good idea to have a better coin selection mechanism, but the Electrum one is not the right approach.

Varunram:
Well, that depends ;  bitcoin core has a very nice selection mechanism (Murch's algorithm) that was merged for the last release, but its overkill for what we're doing at this moment. Electrum is better in some cases than what we have right now (better when your utxos are of similar size) and worse at some (huge differences in utxos size), so its really up to us to decide on an algo.

Varunram:
Closing this for now.

PR #124, opened 2017-12-11
Powless2

adiabat:
This updates powless to use this indexer API:
https://github.com/gertjaap/blockchain-indexer

Looks like it works, and can still be updated more once merged in.  upsv still works and powless can be selected independently for each coin.


adiabat:
No objections for 10 hours; I'm merging :)

PR #126, opened 2017-12-20
Fix Travis

Varunram:
<a href="https://github.com/mit-dci/lit/blob/master/test/test_basic.py#L116">Line 116</a> in test_basic.py still seems to be causing problems. Commented it out for now.


adiabat:
Oh cool it has a green checkmark, that is awesome :)

PR #134, opened 2018-01-08
utxo import / export

adiabat:
This adds the import RPC command which imports a utxo with a private key.  It doesn't import the private key and rescan to find utxos, and as such it doesn't add a new address to watch for payments.

The dump RPC command has been changed to output utxos in this format.  

TODO: (after merging this I guess: add WIF format export alongside this, for compatibility)

### DIFF
--- a/cmd/lit-af/chancmds.go
+++ b/cmd/lit-af/chancmds.go
@@ -201,35 +201,3 @@ func (lc *litAfClient) Push(textArgs []string) error {
 
 	return nil
 }
-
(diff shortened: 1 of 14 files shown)


Varunram:
ACK 7629f8145f4327c9c378a63154238b4297179176 (The logs before removal were extra nice for debugging :) ) I don't know why we get a red cross on the tests though..

Varunram:
I have a rebase branch for this that I use often, I can build on top of the commits here and put in a new PR :)

PR #135, opened 2018-01-09
Disconnect old/invalid spv nodes

Varunram:


adiabat:
Great!  For other coins might be useful to have min needed versions, maybe in coinparams.  This works for bitcoin which is enough for now.

PR #136, opened 2018-01-17
Include data with push command, and provide a history command to view previous payments

metalicjames:
This pull request adds an extra parameter to the channel push command so you can include a 32 byte data payload with the push. This allows recipients so associate a payment with a data payload that is communicated out-of-band. 

It also include a `history` RPC and `lit-af` command to allow users to see the data payload for each push, as well as their relative amounts.

These changes breaks existing databases so you'll need a new one (delete `ln.db`).


adiabat:
Data FTW.

PR #137, opened 2018-01-18
Add data field to fund operation

metalicjames:
Similarly to this PR: https://github.com/mit-dci/lit/pull/136, add a data field to the fund command.

This change does not break existing databases but is not backwards compatible protocol-wise.

### DIFF
--- a/cmd/lit-af/chancmds.go
+++ b/cmd/lit-af/chancmds.go
@@ -112,9 +112,11 @@ func (lc *litAfClient) FundChannel(textArgs []string) error {
 	if len(textArgs) > 4 {
 		data, err := hex.DecodeString(textArgs[4])
 		if err != nil {
-			return err
+			// Wasn't valid hex, copy directly and truncate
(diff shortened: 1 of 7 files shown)


cmd/lit-af/chancmds.go
```diff
@@ -106,6 +108,15 @@ func (lc *litAfClient) FundChannel(textArgs []string) error {
 	if err != nil {
 		return err
 	}
+
+	if len(textArgs) > 4 {
+		data, err := hex.DecodeString(textArgs[4])
+		if err != nil {
+			return err
+		}
+		copy(args.Data[:], data[:32])
```
adiabat: This crashes lit-af, `panic: runtime error: slice bounds out of range`

Can't assume it's 64 hex characters... but also, why assume it's hex at all?  Maybe just look at the len() of textArgs[4] and truncate if it's more than 32?

adiabat:
Ok fixed cool

PR #138, opened 2018-01-18
Use SSL instead of WebSockets

Varunram:


adiabat:
This breaks all the tests, which use websockets.  It'd break them anyway with just the addition of TLS, but the websockets also.  I think it's easier to leave websockets in for now and not change the tests (unless the client also has to do some new stuff to use the certs)

Varunram:
Done!

Varunram:
A hack yfix for python tests (they don't check the SSL cert's authenticity), but should hold until we move completely to tls and have new tests (hopefully).

Varunram:
Moved the entire test suite to TLS in #143.

PR #139, opened 2018-01-29
Fix VTC and VTCTEST parameters for uspv

metalicjames:
This PR fixes Vertcoin and Vertcoin testnet sync when using uspv rather than powless.


adiabat:
OK looks good

PR #141, opened 2018-01-30
Refactored code in dbio.go and fixed wrong hash print when creating header.bin file

frostevent:


.gitignore
```diff
@@ -30,3 +30,6 @@ Session.vim
 .netrwhist
 # auto-generated tag files
 tags
+
+# IntelliJ
+.idea
```
Varunram: Nice catch, should add .DS_Store and similar for atom too. Right now, I honk we have a limited set of stuff in "Editor temp files", maybe we could expand it to `.\*` in the future.

wallit/dbio.go
```diff
@@ -64,35 +64,44 @@ func (w *Wallit) AddPorTxoAdr(kg portxo.KeyGen) error {
...
-// currently returns 20 byte arrays, which
-// can then be converted somewhere else into bech32 addresses (or old base58)
-func (w *Wallit) AdrDump() ([][20]byte, error) {
-	var i, last uint32 // number of addresses made so far
-	var adrSlice [][20]byte
+// getNumberOfKeys retrieves the number of p2pkh keys in the database and check whether
+// the maximum number of keys was reach or not.
+func (w *Wallit) GetNumberOfKeys() (uint32, error) {
+	// Could be placed somewhere else, perhaps it's conf ?
+	var maxKeys = uint32(1 << 20) // the reasonable limit of addresses belonging to one Wallit
```
Varunram: Well, I guess *technically* there's no upper limit on the number of addresses, but this would help :)

frostevent: I haven't found any reasons for why such a limit was enforced. The main point here is to stay coherent between functions checking this limit, and not have magic numbers all around the place.

Varunram: Agreed

.gitignore
frostevent: I usually try to put only what is stricly necessary in the .gitignore file to avoid unexpected git missing file later on.

PR #142, opened 2018-02-01
Basic Link-State channel graph generation

metalicjames:
This PR implements a form of Link-State routing for Lit. Peers send each other information about their open and connected channels periodically and rebroadcast such advertisements from other peers. This allows each Lit node to build its own channel graph. Once multi-hop is implemented, this graph could be used in conjunction with Dijkstra's algorithm to find a route for a payment. 

This also adds the command `graph` to `lit-af` which dumps a Graphviz DOT encoded version of the channel graph. e.g. image 1

### DIFF
--- a/cmd/lit-af/netcmds.go
+++ b/cmd/lit-af/netcmds.go
@@ -31,6 +31,33 @@ var conCommand = &Command{
 	ShortDescription: "Make a connection to another host by connecting to their pubkeyhash\n",
 }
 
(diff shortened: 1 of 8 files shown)


adiabat:
Cool.  I will merge this but that is not an endorsement of the correctness of the structs / messages; some of these things will have to change.  This is just for graphing now though so nothing super serious & bugs won't cause people to lose money or anything.

PR #143, opened 2018-02-08
Port Python tests to SSL

Varunram:
Instead of using websockets over tls, just use tls. Fixed python tests to run with the new setup (it still contains the old names like self.ws though). Replaces #138


alindr:
we need to keep websocket compatibility. we'd use compatibility with webui.

Varunram:
well, this PR was supposed to shift lit to SSL and not only the python tests. Shifting web's compatibility isn't to tough with SSL too.

PR #150, opened 2018-03-03
Add new subdirectory consts

Varunram:
have all constants in consts/consts.go
and import wherever needed


adiabat:
Looks good, thanks!

PR #158, opened 2018-04-24
First version of Discreet Log Contracts

gertjaap:
This first version allows peers to import oracles, create new contracts, offer them to peers, accept/decline, and settle. It also watches the blockchain for settlements by others to claim the correct outputs back to the wallit.

### DIFF
--- a/lnutil/dlclib.go
+++ b/lnutil/dlclib.go
@@ -48,6 +48,7 @@ type DlcContract struct {
 	Status                                   DlcContractStatus                // Status of the contract
 	OurFundingInputs, TheirFundingInputs     []DlcContractFundingInput        // Outpoints used to fund the contract
 	TheirSettlementSignatures                []DlcContractSettlementSignature // Signatures for the settlement transactions
+	FundingOutpoint                          wire.OutPoint                    // The outpoint of the funding TX we want to spend in the settlement - for easier monitoring
 }
 
 // DlcContractDivision describes a single division of the contract. If the oracle predicts OracleValue, we receive ValueOurs
@@ -184,6 +185,9 @@ func DlcContractFromBytes(b []byte) (*DlcContract, error) {
 		copy(c.TheirSettlementSignatures[i].Signature[:], buf.Next(64))
 	}
 
+	copy(op[:], buf.Next(36))
+	c.FundingOutpoint = *OutPointFromBytes(op)
+
 	return c, nil
 }
 
@@ -242,6 +246,9 @@ func (self *DlcContract) Bytes() []byte {
 		wire.WriteVarInt(&buf, 0, uint64(self.TheirSettlementSignatures[i].Outcome))
 		buf.Write(self.TheirSettlementSignatures[i].Signature[:])
 	}
+
+	opArr := OutPointToBytes(self.FundingOutpoint)
+	buf.Write(opArr[:])
 
 	return buf.Bytes()
 }
@@ -404,13 +411,13 @@ func computePubKey(pubA, pubR [33]byte, msg []byte) ([33]byte, error) {
 }
 
 // SettlementTx returns the transaction to settle the contract. ours = the one we generate & sign. Theirs (ours = false) = the one they generated, so we can use their sigs
-func SettlementTx(c *DlcContract, d DlcContractDivision, contractInput wire.OutPoint, ours bool) (*wire.MsgTx, error) {
+func SettlementTx(c *DlcContract, d DlcContractDivision, ours bool) (*wire.MsgTx, error) {
 
 	tx := wire.NewMsgTx()
 	// set version 2, for op_csv
 	tx.Version = 2
 
-	tx.AddTxIn(wire.NewTxIn(&contractInput, nil, nil))
+	tx.AddTxIn(wire.NewTxIn(&c.FundingOutpoint, nil, nil))
 
 	totalFee := int64(1000) // TODO: Calculate
 	feeEach := int64(float64(totalFee) / float64(2))
--- a/lnutil/dlclib.go
+++ b/lnutil/dlclib.go
@@ -5,6 +5,11 @@ import (
 	"bytes"
 	"encoding/binary"
 	"fmt"
+	"math/big"
+
+	"github.com/adiabat/btcd/btcec"
+	"github.com/adiabat/btcd/chaincfg/chainhash"
+	"github.com/adiabat/btcutil"
 
 	"github.com/adiabat/btcd/txscript"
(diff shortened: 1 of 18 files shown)


qln/dlc.go
```diff
@@ -0,0 +1,611 @@
...
+		return fmt.Errorf("You cannot offer a contract to someone that is not in draft stage")
+	}
+
+	if !nd.ConnectedToPeer(peerIdx) {
+		return fmt.Errorf("You are not connected to peer %d, do that first", peerIdx)
+	}
+
+	var nullBytes [33]byte
+	// Check if everything's set
+	if bytes.Equal(nullBytes[:], c.OracleA[:]) {
```
adiabat: nitpicky but with arrays you can just say
``` if c.OracleA == nullBytes {``` which is shorter / easier

File: lnutil/dlclib.go [hunk 2]
```diff
@@ -0,0 +1,481 @@
...
+)
+
+// scalarSize is the size of an encoded big endian scalar.
+const scalarSize = 32
+
+// DlcContract is a struct containing all elements to work with a Discreet Log Contract. This struct is stored in the database
+type DlcContract struct {
+	Idx                                      uint64                           // Index of the contract for referencing in commands
+	PeerIdx                                  uint32                           // Index of the peer we've offered the contract to or received the contract from
+	PubKey                                   [33]byte                         // Key of the contract
```
adiabat: PubKey is ambiguous here; I'm not sure what key this is

gertjaap: Pubkey is used by remote peers to identify the contract - they don't know the contract's index in our database. Hence if a peer is accepting my contract offer, i need to know which contract he is referring to.

lnutil/dlclib.go
```diff
@@ -0,0 +1,481 @@
...
+	CoinType                                 uint32                           // Coin type
+	OracleA, OracleR                         [33]byte                         // Pub keys of the oracle
+	OracleTimestamp                          uint64                           // The time we expect the oracle to publish
+	Division                                 []DlcContractDivision            // The payout specification
+	OurFundingAmount, TheirFundingAmount     int64                            // The amounts either side are funding
+	OurChangePKH, TheirChangePKH             [20]byte                         // PKH to which the contracts funding change should go
+	OurFundMultisigPub, TheirFundMultisigPub [33]byte                         // Pubkey used in the funding multisig output
+	OurPayoutPub, TheirPayoutPub             [33]byte                         // Pubkey to which the contracts are supposed to pay out
```
adiabat: probably makes sense to also add a payout pubkey hash, and perhaps rename PayoutPub to PayoutBase (as it will not be used directly as a public key)

lnutil/dlclib.go
```diff
@@ -0,0 +1,481 @@
...
+
+	return wire.NewTxOut(value, scriptBytes)
+}
+
+// DLC Commit script makes a script that pays to (PubKeyPeer+PubKeyOracleSig or (OurPubKey and TimeDelay))
+// We send this over (signed) to the other side. If they publish the TX with the correct script they can use
+// The oracle's signature and their own private key to claim the funds from the output. However,
+// If they send the wrong one, they won't be able to claim the funds - and we can claim them once the
+// Time delay has passed.
+func DlcCommitScript(pubKeyPeer, pubKeyOracleSig, ourPubKey [33]byte, delay uint16) []byte {
```
adiabat: This is almost the same as CommitScript() in lnlib with, so clearer to call CombinePubs() and then CommitScript() which would make this function ~2 lines

lnutil/dlclib.go
```diff
@@ -0,0 +1,481 @@
...
+	builder.AddOp(txscript.OP_CHECKSIG)
+
+	// never any errors we care about here.
+	s, _ := builder.Script()
+
+	return s
+}
+
+// BigIntToEncodedBytes converts a big integer into its corresponding
+// 32 byte little endian representation.
```
adiabat: sure this is little endian? I think it's big..?

adiabat: Also it's quite possible that *I* wrote this and it was copied here (I sortof vaguely remember writing something like this to pad out s values) so I might have written that it's little endian instead of big.  Pretty sure bigint.Bytes() returns big endian though.

gertjaap: I copied it from https://github.com/adiabat/dlctest/blob/master/sig.go#L45

Varunram: yup, big endian (ref. https://golang.org/pkg/math/big/#Int.Bytes)

lnutil/dlclib.go
```diff
@@ -0,0 +1,481 @@
...
+
+	// Ours = the one we generate & sign. Theirs (ours = false) = the one they generated, so we can use their sigs
+	if ours {
+		if valueTheirs > 0 {
+			tx.AddTxOut(DlcOutput(c.TheirPayoutPub, oracleSigPub, c.OurPayoutPub, valueTheirs))
+		}
+
+		if valueOurs > 0 {
+			var ourPayoutPKH [20]byte
+			copy(ourPayoutPKH[:], btcutil.Hash160(c.OurPayoutPub[:]))
```
adiabat: If the PKH only output uses the same pubkey as in the DlcOutput script, that can allow someone to figure out what oracle signature is being used by observing the transaction (actually not this one; they will have to wait until the PKH output is spent, and then they'll be able to figure it out)

So we need to use a separate PayoutPKH.  In lit channels the whole pubkey is given: MyRefundPub and TheirRefundPub.  I don't think you need the actual pubkey, so maybe pubkey hash alone is OK.  Or the whole key, either way works.

adiabat: Actually I think this attack only can occur if both parties broadcast their SettlementTx at the same time, as that would be needed to reveal both my pubkey and the pubkey+oracleKey.

still, that's possible, so it's better to have a separate PKH output key.

qln/dlc.go
```diff
@@ -0,0 +1,611 @@
...
+
+	if bytes.Equal(nullBytes[:], c.OracleR[:]) {
+		return fmt.Errorf("You need to set an R-point for the contract before offering it")
+	}
+
+	if c.OracleTimestamp == 0 {
+		return fmt.Errorf("You need to set a settlement time for the contract before offering it")
+	}
+
+	if c.CoinType == 0 {
```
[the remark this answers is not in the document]
gertjaap: Added a new constant COINTYPE_NOT_SET with ^uint32(0) (maximum UINT32) as value. I think we won't have that many coins eventually, so this will probably be a safe value to use.

PR #162, opened 2018-05-21
dlc workflow rework

Varunram:
PLEASE NOTE: intended to be merged after #161 OR intended to be merged instead of #161 provided the proposed changes are acceptable.

This combines all the commands for which channel ID is a common parameter and adds an argument called `setparams` . Now you can create a contract with minimal commands (`new` and `setparams`). Parsing on the UI side should be easy since this is just one command and changes should be minimal. If this PR is accepted, I shall port the changes over to the UI. Same goes with the docs.

An advantage of merging all commands is that you can create contracts with a single command without passing in the channel id to each command. Disadvantage is that you lose granularity and would potentially have to run `setparams` each time you want to change a parameter.

Other commands (`setoracle`, `settime`, etc) can also simultaneously exist along with `setparams` although this would just be duplicating most of the code. But this can be done and should be easy to do, if needed.

P.S. this builds upon gertjaap's `dlcfwd` branch, so the first 7 commits aren't related to changes. Commits https://github.com/mit-dci/lit/pull/162/commits/a9ff4cbc1bbfe03311e256451b7bec6db94cbab5 and https://github.com/mit-dci/lit/pull/162/commits/0c9c06571aeb07109004a5394367d91baa73c651 contain the proposed modifications.


gertjaap:
@Varunram wondering how relevant this PR still is. I think it *might* be interesting to have a single setparams command, although i am not really a fan of its user unfriendlyness. If we were to add such a command, i'd prefer to be able to set all parameters at once, and add it on top of the separate commands.

Let me know what you think.

Varunram:
Closing this with regard to the discussions we had the other day.

PR #164, opened 2018-05-29
Replace lndc by brontide (from lnd) to support BOLT #8.

yaslama:
In order to have interoperability with other LN implementations, the communication between nodes needs to use the protocol described in [BOLT #8](https://github.com/lightningnetwork/lightning-rfc/blob/master/08-transport.md). The protocol is implemented by the brontide package from lnd, which I copied (and modified a little bit to be self contained).
lit to lit communication is not affected by this change, but now lit can communicate with other BOLT compliant LN implementations. Other BOLTs need to be implemented, but this is the first step.


metalicjames:
I don't think it makes much sense to do that since the channel protocol is also completey different. At this point Lit doesn't plan to conform to the BOLT spec as we're persuing a slightly different feature set.

kingonly:
Is the ability to connect to other LN implementations not important to you guys? Can you please clarify what exactly do you mean by "slightly different feature set"?

adiabat:
Thanks!  The brontide package is useful for this and is a good replacement for lndc, which is old and a bit of a mess.

@Varunram  & @yaslama: this won't allow lit nodes to talk to lnd nodes as they have different messages and protocols.  Even implementing the base protocol init messages wouldn't do much. 

@kingonly: Compatibility with other implementations is not something we're focused on.

metalicjames:
okay to test

metalicjames:
ok to test

Varunram:
needs rebase

Varunram:
Rebased this on top of master. If you'd be happy to confirm the changes at #227, that'd be great. If you pull that branch unto this and feel that PR be closed, I'd be happy to close do that too, thanks!

gertjaap:
Since this is rebased in #227 we can close this one.

PR #165, opened 2018-05-29
Fix bug where non-witness outputs were selected in PickUtxos

metalicjames:
If there were a mixture of witness and non-witness outputs in the db then PickUtxos would sometimes select non-witness inputs even when only-witness was set. This prevented channel funding in a number of cases.


adiabat:
Nice catch! looks good

PR #166, opened 2018-05-29
keep reconnecting in case we connect to spam nodes

### LINKED ISSUE #133: Testnet seed connects to Bitcoin cash node
Something very weird happened when  connecting to Peter Todd's testnet seed at `testnet-seed.bitcoin.petertodd.org`:

```
2018/01/04 17:14:33.014980 DB height 1257406
2018/01/04 17:14:33.015010 DB height 1257406
2018/01/04 17:14:33.015041 opened header file /root/.lit/testnet3/header.bin
2018/01/04 17:14:33.015050 Attempting connection based on DNS seed
2018/01/04 17:14:33.046706 Got 38 IPs from DNS seed testnet-seed.bitcoin.petertodd.org
2018/01/04 17:14:33.046826 Attempting connection to node at 64.52.84.86:18333
2018/01/04 17:14:33.255641 wrote 134 byte version message to 64.52.84.86:18333
2018/01/04 17:14:33.462726 got 137 byte response 7f1101002500000000000000a1134e5a00000000000000000000000000000000000000000000ffff31ce73462e812500000000000000000000000000000000000000000000000000af289037bf2824771b2f426974636f696e204142433a302e31362e31284542382e30292f5f51120001
 command: version
2018/01/04 17:14:33.462798 connected to /Bitcoin ABC:0.16.1(EB8.0)/
2018/01/04 17:14:33.462837 remote reports version 1117f (dec 70015)
2018/01/04 17:14:33.462985 got header tip height 1257406
2018/01/04 17:14:33.463209 Got verack.  Whatever.
2018/01/04 17:14:33.463304 Got unknown message type alert
2018/01/04 17:14:33.463569 get headers message has 19 header hashes, first one is 0000000000000168ae0daf3966778af563af35440f0f4de39cfe53d0caa94868
2018/01/04 17:14:33.726934 Got unknown message type sendheaders
2018/01/04 17:14:33.727227 got 1 addresses.
2018/01/04 17:14:34.377146 got 2000 headers. Range:
00000000000a3abd905d7ff94fdf9eb754f1b8587b4892937f9b88a3089621ab - 000000000003cfc8793011c3c96af8962c5a7bfba9c1e36d64ba6ce5ebcac151
2018/01/04 17:14:34.393777 header file position: 18017120
2018/01/04 17:14:34.393840 made 2017 header slice
2018/01/04 17:14:34.412326 Header 00000000000a3abd905d7ff94fdf9eb754f1b8587b4892937f9b88a3089621ab attaches at height 1032192
2018/01/04 17:14:34.412396 Header error: reorg message up to height 1034192, but have up to 1257406
2018/01/04 17:14:34.412416 blockTip to 1257406 headerTip 1257406
2018/01/04 17:14:34.412424 no blocks to request, entering wait state
```
This issue might be a good chance to add a couple more testnet seeds from Core (even Peter's seed has changed URLs)

Varunram:
in the case we do connect to bad nodes, disconnect and keep trying. Closes #133 

Basically the same stuff we do above, but can't reduce code repetition due to scope.
EDIT: added three new functions, no repetition.

### DIFF
--- a/uspv/init.go
+++ b/uspv/init.go
@@ -137,7 +137,7 @@ func (s *SPVCon) Connect(remoteNode string) error {
 		if !strings.Contains(remoteNode, ":") {
 			remoteNode = remoteNode + ":" + s.Param.DefaultPort
 		}
-		listOfNodes = append(listOfNodes, remoteNode)
+		listOfNodes = []string{remoteNode}
 	}
 
 	for len(listOfNodes) != 0 {
@@ -148,6 +148,10 @@ func (s *SPVCon) Connect(remoteNode string) error {
 		}
 		err = s.Handshake(listOfNodes)
 		if err != nil {
+			if len(listOfNodes) == 1 { // when the list is empty, error out
+				return fmt.Errorf("Couldn't establish connection with remote node. Exiting.")
+				break
+			}
 			// means we either have a sapm node or didn't get a resonse. So we Try again
 			log.Println(err)
 			log.Println("Couldn't establish connection with node. Proceeding to the next one")
--- a/uspv/init.go
+++ b/uspv/init.go
@@ -13,7 +13,7 @@ import (
 	"github.com/mit-dci/lit/lnutil"
 )
 
-func GetSeedAdrs(s *SPVCon, remoteNode string) error {
+func (s *SPVCon) GetSeedAdrs(remoteNode string) error {
 	var err error
 	// slice of IP addrs returned from the DNS seed
 	var seedAdrs []string
@@ -48,7 +48,7 @@ func GetSeedAdrs(s *SPVCon, remoteNode string) error {
 				}
 			}
 
-			err = ConnectToSeedAdrs(s, seedAdrs)
+			err = s.ConnectToSeedAdrs(seedAdrs)
 			if err != nil {
 				if i != len(s.Param.DNSSeeds)-1 {
 					continue
@@ -69,14 +69,14 @@ func GetSeedAdrs(s *SPVCon, remoteNode string) error {
 		if err != nil {
 			return err
 		}
-		err := Handshake(s, nil)
+		err := s.Handshake(nil)
 		if err != nil {
 			return err
 		}
 	}
 	return nil
 }
-func ConnectToSeedAdrs(s *SPVCon, seedAdrs []string) error {
+func (s *SPVCon) ConnectToSeedAdrs( seedAdrs []string) error {
 	// now have some IPs, go through and try to connect to one.
 	var err error
 	var connected bool
@@ -94,7 +94,7 @@ func ConnectToSeedAdrs(s *SPVCon, seedAdrs []string) error {
 			connected = true
 		}
 		if connected {
-			err := Handshake(s, seedAdrs)
+			err := s.Handshake(seedAdrs)
 			if err != nil {
 				continue
 			}
@@ -104,7 +104,7 @@ func ConnectToSeedAdrs(s *SPVCon, seedAdrs []string) error {
 	return nil
 }
 
-func Handshake(s *SPVCon, seedAdrs []string) error {
+func (s *SPVCon) Handshake(seedAdrs []string) error {
 	// assign version bits for local node
 	s.localVersion = VERSION
 	myMsgVer, err := wire.NewMsgVersionFromConn(s.con, 0, 0)
@@ -151,7 +151,7 @@ func Handshake(s *SPVCon, seedAdrs []string) error {
 	if !(strings.Contains(mv.UserAgent, "Satoshi") || strings.Contains(mv.UserAgent, "btcd")) && (len(seedAdrs) != 0) {
 		// TODO: improve this filtering criterion
 		seedAdrs = seedAdrs[1:] // delete the one we tried with
-		err = ConnectToSeedAdrs(s, seedAdrs)
+		err = s.ConnectToSeedAdrs(seedAdrs)
 		if err != nil {
 			return fmt.Errorf("Couldn't connect to any seed!")
 		}
@@ -178,7 +178,7 @@ func Handshake(s *SPVCon, seedAdrs []string) error {
 
 // Connect dials out and connects to full nodes.
 func (s *SPVCon) Connect(remoteNode string) error {
-	err := GetSeedAdrs(s, remoteNode)
+	err := s.GetSeedAdrs(remoteNode)
 	if err != nil {
 		return fmt.Errorf("Couldn't connect to the node you provided!")
 	}
--- a/uspv/init.go
+++ b/uspv/init.go
@@ -13,74 +13,32 @@ import (
 	"github.com/mit-dci/lit/lnutil"
 )
 
-func (s *SPVCon) GetSeedAdrs(remoteNode string) error {
+func (s *SPVCon) GetSeedAdrs(seed string) ([]string, error) {
 	var err error
-	// slice of IP addrs returned from the DNS seed
-	var seedAdrs []string
-	// if remoteNode is "yes" but no IP specified, use DNS seed
-	if lnutil.YupString(remoteNode) {
-		// connect to DNS seed based node
-		log.Printf("Attempting connection based on DNS seed\n")
-		if len(s.Param.DNSSeeds) == 0 {
-			// no available DNS seeds
-			return fmt.Errorf(
-				"Can't connect: No known DNS seeds for %s", s.Param.Name)
-		}
+	var seedAdrs []string // slice of IP addrs returned from the DNS seed
+	log.Printf("Attempting to retrieve peers to connect to based on DNS Seed\n")
 
-		// go through seeds to get a list of IP addrs
-		for i, seed := range s.Param.DNSSeeds {
-			seedAdrs, err = net.LookupHost(seed)
-			if err == nil {
-				// got em
-				log.Printf("Got %d IPs from DNS seed %s\n", len(seedAdrs), seed)
-			} else {
-				// gotta keep trying for those IPs
-				continue
-			}
-
-			if len(seedAdrs) == 0 {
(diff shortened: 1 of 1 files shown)


uspv/init.go
```diff
@@ -13,10 +13,8 @@ import (
 	"github.com/mit-dci/lit/lnutil"
 )
 
-// Connect dials out and connects to full nodes.
-func (s *SPVCon) Connect(remoteNode string) error {
+func GetSeedAdrs(s *SPVCon, remoteNode string) error {
```
adiabat: This should be a method on SPVCon; so should the others

Varunram: will do

uspv/init.go
```diff
@@ -72,8 +69,42 @@ func (s *SPVCon) Connect(remoteNode string) error {
...
+			if err != nil {
+				continue
+			}
+			break
+		}
 	}
+	return nil
+}
 
+func Handshake(s *SPVCon, seedAdrs []string) error {
```
adiabat: Here too, the way most of these things are organized is there's a big persistent struct with a bunch of state and methods on it.  (whether that's the best way to do it or not is another question :)  But might as well stick with it consistently)

Varunram: done, are the names fine?

uspv/init.go
```diff
@@ -137,6 +173,15 @@ func (s *SPVCon) Connect(remoteNode string) error {
 		return err
 	}
 	s.WBytes += uint64(n)
+	return nil
+}
+
+// Connect dials out and connects to full nodes.
+func (s *SPVCon) Connect(remoteNode string) error {
```
adiabat: I think this should be organized so that more is going on in Connect().  Connect can call all the sub-functions directly

uspv/init.go
```diff
@@ -137,6 +173,15 @@ func (s *SPVCon) Connect(remoteNode string) error {
...
+	err := s.GetSeedAdrs(remoteNode)
```
adiabat: Here, GetSeedAdrs() takes remoteNode, but logically it doesn't need a remote node as an argument, because the point of getting the seed addresses is to figure out a remote node.

If GetSeedAdrs() just returns a list of adrs (which sounds like what it would do) then it doesn't need an argument.

uspv/init.go
```diff
@@ -72,8 +69,42 @@ func (s *SPVCon) Connect(remoteNode string) error {
...
 			return err
 		}
+		err := s.Handshake(nil)
+		if err != nil {
+			return err
+		}
+	}
+	return nil
+}
+func (s *SPVCon) ConnectToSeedAdrs( seedAdrs []string) error {
```
adiabat: This can be reworked into a DialNode() function; takes in a [single / slice of] addresses and tries to connect, assigning that connection to s.con.

It can just return an error, and doesn't need to call any sub functions; after it returns (without error) there will be an s.con that's waiting for a handshake.

uspv/init.go
```diff
@@ -117,9 +148,13 @@ func (s *SPVCon) Connect(remoteNode string) error {
 		return fmt.Errorf("Remote node version: %x too old, exiting.", mv.ProtocolVersion)
 	}
 
-	if strings.Contains(mv.UserAgent, "ABC") {
-		// if we connected through a DNS Peer and it doesn't implement service bit filtering
-		return fmt.Errorf("Remote node %s invalid", mv.UserAgent)
+	if !(strings.Contains(mv.UserAgent, "Satoshi") || strings.Contains(mv.UserAgent, "btcd")) && (len(seedAdrs) != 0) {
```
adiabat: whitelists are ugly and we should filter based on behavior, not based on names, but we can leave this for now until the connection management is more built-out.

uspv/init.go
```diff
@@ -13,67 +13,48 @@ import (
 	"github.com/mit-dci/lit/lnutil"
 )
 
-// Connect dials out and connects to full nodes.
-func (s *SPVCon) Connect(remoteNode string) error {
+func (s *SPVCon) GetSeedAdrs(seed string) ([]string, error) {
```
adiabat: as it no longer uses SPVCon, it doesn't need to be a method

uspv/init.go
```diff
@@ -13,67 +13,48 @@ import (
...
+	}
+	return seedAdrs, nil
+	// return the seedAdrs here
+}
 
-		// open TCP connection to specified host
-		s.con, err = net.Dial("tcp", remoteNode)
-		if err != nil {
-			return err
+func (s *SPVCon) DialNode(seedAdrs []string) error {
```
adiabat: Doesn't (or shouldn't) need the argument here

uspv/init.go
Varunram: nice catch, fixing.

uspv/init.go
adiabat: len(seedAdrs) isn't relevant here; the handshake works or doesn't regardless

uspv/init.go
adiabat: Nevermind, this is needed -- instead maybe change the name, as it's not a seedAdrs slice, since it might not be from a seed.  Just nodeAdrs?

uspv/init.go
```diff
@@ -137,6 +119,54 @@ func (s *SPVCon) Connect(remoteNode string) error {
...
+			err = s.Handshake(seedAdrs)
+			if err != nil {
+				// means we either have a sapm node or didn't get a resonse. So we Try again
+				log.Println(err)
+				log.Println("Couldn't establish connection with node. Proceeding to the next one")
+				continue
+			}
+			break
+		}
+	} else { // else connect to user-specified node
```
adiabat: we don't need a separate codepath for a user-supplied address.  If we create a address slice with only the supplied address, the rest of the code should work the same, but with skipping only GetSeedAdrs()

uspv/init.go
Varunram: yeah cool

uspv/init.go
```diff
@@ -137,7 +120,42 @@ func (s *SPVCon) Connect(remoteNode string) error {
...
+		// if remoteNode is "yes" but no IP specified, use DNS seed
+		listOfNodes, err = s.GetListOfNodes()
+		if err != nil {
+			return err
+		}
+	} else { // else connect to user-specified node
+		if !strings.Contains(remoteNode, ":") {
+			remoteNode = remoteNode + ":" + s.Param.DefaultPort
+		}
+		listOfNodes = append(listOfNodes, remoteNode)
```
adiabat: Does the same thing, but `listOfNodes = []string{remoteNode}` looks nicer.

uspv/init.go
```diff
@@ -137,7 +120,42 @@ func (s *SPVCon) Connect(remoteNode string) error {
...
+		}
+		err = s.Handshake(listOfNodes)
+		if err != nil {
+			// means we either have a sapm node or didn't get a resonse. So we Try again
+			log.Println(err)
+			log.Println("Couldn't establish connection with node. Proceeding to the next one")
+			listOfNodes = listOfNodes[1:]
+			continue
+		}
+		break
```
adiabat: Need an error message for when it runs out of nodes to connect to, and still hasn't done a successful handshake.

uspv/init.go
Varunram: yeah, looks nicer.

uspv/init.go
Varunram: thanks!

Varunram:
Incorporated more changes, hope its all good now.

uspv/init.go
```diff
@@ -137,7 +120,46 @@ func (s *SPVCon) Connect(remoteNode string) error {
...
 
+	for len(listOfNodes) != 0 {
+		err = s.DialNode(listOfNodes)
+		if err != nil {
+			listOfNodes = listOfNodes[1:]
+			continue
+		}
+		err = s.Handshake(listOfNodes)
+		if err != nil {
+			if len(listOfNodes) == 1 { // when the list is empty, error out
```
adiabat: remember whether the DNS seed lookup was performed or not.  Users should get an error indicating the host they supplied is down, vs DNS not working.

uspv/init.go
```diff
@@ -137,7 +120,46 @@ func (s *SPVCon) Connect(remoteNode string) error {
...
+	} else { // else connect to user-specified node
+		if !strings.Contains(remoteNode, ":") {
+			remoteNode = remoteNode + ":" + s.Param.DefaultPort
+		}
+		listOfNodes = []string{remoteNode}
+	}
 
+	for len(listOfNodes) != 0 {
+		err = s.DialNode(listOfNodes)
+		if err != nil {
```
adiabat: display something here so the user can know what's going on; this could be the address they supplied failing, or lots of DNS seed hosts not answering and could take a while

Varunram: cool

uspv/init.go
adiabat: Also, even if it breaks, the function continues to start an incomingMessageHandler() which will won't do anything.  Instead Connect() needs to return an error.

Varunram: oh, the break there is unnecessary since there's a return above. Deleting that.

Varunram:
test failing is unrelated, it fails due to the socket exception thing we had before.

metalicjames:
retest this please

metalicjames:
restest this please

Varunram:
@metalicjames I think a rebase should work, this branch was done before your PR, will rebase and push

Varunram:
retest this please

Varunram:
Okay, looks like its the weird socket error thing again, not to do with `go build` or related stuff.

metalicjames:
It only seems to happen with this PR though. The master branch and other PRs have tested reliably.

adiabat:
OK, looks good, merging this.  A good next step would be to detect when a host disconnects from us, and then try to connect to another host and resume operation.

PR #167, opened 2018-05-29
Fix race condition in channel push code

metalicjames:
The justice transactions were being built in goroutines using the shared channel state. This created a race condition where the justice tx code and subsequent push sometimes stepped on each others' toes. 

Additionally in a couple of the message handlers the `LitNode` was being passed by-value, and so the mutex contained within was being copied. This meant the actual lock was not being locked. 

This PR seems to make channels stable over latent connections. I was able to do 10,000 consecutive pushes without issue, whereas beforehand I could only make it to ~50-300 before failure.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,6 @@
 # lit - a lightning node you can run on your own
 ![Lit Logo](litlogo145.png)
+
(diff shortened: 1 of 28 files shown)


adiabat:
awesome, thanks!

PR #170, opened 2018-05-30
Fix a bunch of compile errors when running go tests

metalicjames:
Mostly problems with misuse of `Printf` and `Println`. This PR is required to make the tests/build pass.


metalicjames:
retest this please

PR #171, opened 2018-06-01
Re-register refund addresses with wallit after restarting lit

metalicjames:
When a channel is first funded your refund address is registered with wallit so the resulting UTXO is picked up by a close. After Lit was restarted we were not re-registering those refund addresses with the wallit. As a result, closing a channel resulted in you receiving no new UTXO.

### DIFF
--- a/qln/init.go
+++ b/qln/init.go
@@ -2,7 +2,6 @@ package qln
 
 import (
 	"fmt"
-	"log"
 	"path/filepath"
 
 	"github.com/adiabat/btcutil"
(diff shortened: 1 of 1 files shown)


metalicjames:
test this please

metalicjames:
ok to test

adiabat:
Looks good, seems to work.  Big oversight not registering those, oops!  Though if there were anyone with money not appearing due to this bug, upgrading via this PR will re-discover the funds.

PR #173, opened 2018-06-04
Add a timeout to RPC calls in lit-af of 10 seconds

metalicjames:
Previously failed calls would hang forever. Now they timeout after 10 seconds.


metalicjames:
retest this please

adiabat:
Heh makes sense.  Though eventually we need to change a bunch more of the RPC stuff, this is a good fix for now :)

PR #179, opened 2018-06-08
Fix race condition in pushpull code to allow colliding pushes

metalicjames:
Some handlers would run while others were in-flight meaning that the channel state got messed up. Add an extra mutex to `Qchan` to ensure only one message handler runs at a time. 

This also means rather than erroring out when the channel is busy, we can now make colliding pushes that resolve simultanous pushes.

This PR also adds a bunch of useful logging to the pushpull code.

### DIFF
--- a/qln/lnchannels.go
+++ b/qln/lnchannels.go
@@ -3,6 +3,7 @@ package qln
 import (
 	"bytes"
 	"fmt"
+	"sync"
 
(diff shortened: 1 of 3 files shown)


adiabat:
Cool - long term we should refactor push RPC call but this fixes collisions - cool

PR #180, opened 2018-06-08
use insight api instead of blockexplorer

Varunram:
block explorer is basically a fork of insight's API, block explorer is not well maintained now, although insight is, so we can shift to insight with the same endpoints.


adiabat:
OK cool; later it should support a bunch of different APIs that could be runtime arguments.  But if this works for now, sounds good.

PR #184, opened 2018-06-13
Increase max message size to 16MB (needed for DLC)

gertjaap:
This was not yet merged into master. It's needed for DLC's with larger numbers of signatures.


Varunram:
I guess this made it into `dlcfwd` too, can see the change over there

gertjaap:
@Varunram Yes, they were part of the DLCFWD. But it makes sense to merge this separately, since it's not specific to DLC - just a prerequisite. It might have other uses.

adiabat:
This is in keeping with satoshi's vision of > 65KB messages :)

PR #187, opened 2018-06-13
Updates to WALKTHROUGH.md.

delbonis:
Updated to use the new address format, command responses (like `con`), and explanation for support cointypes that was added since it was originally written.

### DIFF
--- a/WALKTHROUGH.md
+++ b/WALKTHROUGH.md
@@ -1,5 +1,7 @@
 2017-01-09 Lit sortof works on testnet.  There are known bugs / omissions / errors, and also unknown ones!  Reporting bugs / crashes is helpful, and fixing them is even more helpful.
 
+2018-06-12 Updated to use new addresses and command formats.  Lit also works much better than it did in the above note.
+
 ## Lit 0.0 Walkthrough
 
 This walkthrough is to set people up who want to send payments over channels on a test network. If you haven't already, make sure to go to the [README](./README.md) for setup instructions.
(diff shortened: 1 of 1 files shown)


delbonis:
I ended up making a new GitHub account anyways so someone can remove @treyzania from the org and add this once instead.

WALKTHROUGH.md
```diff
@@ -100,22 +99,21 @@ Bob puts the pubkey@hostname for Alice and connects.  Then he says hi to Alice.
...
 
 Bob is connected to Alice and wants to open a payment channel. If he has enough (segwit) money and is connected to Alice, he can open a channel.
 
 ```
-lit-af# fund 1 50000000 0
+lit-af# fund 1 1 50000000 0
 ```
 
-This opens a channel with peer 1 (Alice) with a channel capacity of 50,000,000 satoshis (half a coin), and sends 0 satoshis over in the creation process.  Bob starts out with all 50,000,000 satoshis in the channel, so only he can send to Alice.
+This opens a channel with peer 1 (Alice) with a channel capacity of 50,000,000 satoshis (half a coin), and sends 0 satoshis over in the creation process.  Bob starts out with all 50,000,000 satoshis in the channel, so only he can send to Alice.  The second "1" is to specify the cointype, which is 1 for the Bitcoin Testnet3.
```
Varunram: s/for the/for should be better :)

delbonis: Fixed.

adiabat:
Looks good, thanks for the update

PR #188, opened 2018-06-13
Add Makefile and env.sh

Varunram:
For all GOPATH issues


adiabat:
OK, I generally suggest just going with the gopath flow, but if you need to put things in a different place this works.

PR #189, opened 2018-06-13
Added list of cointypes in help command.

delbonis:
I'm not sure what everyone thinks of the formatting for the cointypes, of course we can always change it.


Varunram:
test this please

Varunram:
retest this please

adiabat:
great -- I think we can remove all cointypes from UI (even text UI) and replace with bech32 prefixes, then we won't have to deal with these numbers

PR #202, opened 2018-06-14
Optional SOCKS5 for LNDC connections

metalicjames:
With this PR you can specify a SOCKS5 proxy with the --proxy option to Lit. Currently connections to full nodes and powless APIs still take place via the normal network. When the proxy is enabled Lit does not attempt to announce itself on the tracker and uses the proxy for tracker lookups.


adiabat:
Looks good, now tor / ssh -D will be easy to use

delbonis:
that was fast

Varunram:
lightning hands

PR #204, opened 2018-06-15
[trivial] edit log messages

Varunram:
Delete one rogue `log.Println`message and modify another one. In most cases, the log reads 
```
DB height 1325523
DB height 1325523
```
and in the event of resync, something like 
```
DB height 1325523
DB height 1300000
```
which is somewhat weird.


metalicjames:
We should change all of the `fmt.Printf` calls to `log.Printf`. Right now there is a mix and it's annoying. Everything should just use `log` to maintain formatting and include a timestamp.

Varunram:
hm, good point, maybe I'll do that conversion along with this one

PR #205, opened 2018-06-15
Reconnect to other nodes if we disconnect from one

Varunram:
Really, we must `DialNode` in case we disconnect, but there's the problem of peers being stale when we try to reconnect, so might as well gather `listofNodes` again.

### DIFF
--- a/uspv/msghandler.go
+++ b/uspv/msghandler.go
@@ -13,7 +13,8 @@ func (s *SPVCon) incomingMessageHandler() {
 		n, xm, _, err := wire.ReadMessageWithEncodingN(s.con, s.localVersion,
 			wire.BitcoinNet(s.Param.NetMagicBytes), wire.LatestEncoding)
 		if err != nil {
-			log.Printf("ReadMessageWithEncodingN error.  Disconnecting: %s\n", err.Error())
+			log.Printf("ReadMessageWithEncodingN error.  Disconnecting from given peer. %s\n", err.Error())
+			s.Connect("yes") // really any YupString here
 			return
 		}
 		s.RBytes += uint64(n)
--- a/uspv/msghandler.go
+++ b/uspv/msghandler.go
@@ -16,6 +16,8 @@ func (s *SPVCon) incomingMessageHandler() {
 			log.Printf("ReadMessageWithEncodingN error.  Disconnecting from given peer. %s\n", err.Error())
 			if s.randomNodesOK { // if user wants to connect to localhost, let him do so
 				s.Connect("yes") // really any YupString here
+			} else {
+				s.con.Close()
 			}
 			return
 		}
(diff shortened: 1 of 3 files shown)


adiabat:
Looks good -- doing another DNS seed lookup each time is a bit inefficient but OK for now.  We should store node lists on disk from getaddr & connect to multiple simultaneous clients eventually.

uspv/msghandler.go
```diff
@@ -13,7 +13,8 @@ func (s *SPVCon) incomingMessageHandler() {
 		n, xm, _, err := wire.ReadMessageWithEncodingN(s.con, s.localVersion,
 			wire.BitcoinNet(s.Param.NetMagicBytes), wire.LatestEncoding)
 		if err != nil {
-			log.Printf("ReadMessageWithEncodingN error.  Disconnecting: %s\n", err.Error())
+			log.Printf("ReadMessageWithEncodingN error.  Disconnecting from given peer. %s\n", err.Error())
+			s.Connect("yes") // really any YupString here
```
adiabat: I think we need a check here.  If someone starts with `--tn3 127.0.0.1` they really only want it to connect to localhost.  This would mean that if the connection to localhost failed (after getting established) their node would attempt to connect to random nodes, without the user knowing.

If the user supplies a fullnode, and that node doesn't work, the program's response should be to fail & inform the user, not try other nodes.

Varunram: yeah, I was wondering what to do about it too, I guess I'll  whitelist `127.0.0.1` and `localhost` for now (until a better mechanism)

uspv/msghandler.go
```diff
@@ -13,7 +13,10 @@ func (s *SPVCon) incomingMessageHandler() {
 		n, xm, _, err := wire.ReadMessageWithEncodingN(s.con, s.localVersion,
 			wire.BitcoinNet(s.Param.NetMagicBytes), wire.LatestEncoding)
 		if err != nil {
-			log.Printf("ReadMessageWithEncodingN error.  Disconnecting: %s\n", err.Error())
+			log.Printf("ReadMessageWithEncodingN error.  Disconnecting from given peer. %s\n", err.Error())
```
adiabat: this happens even if the peer isn't given, so maybe just say "disconnecting"
Except it may not even disconnect properly, so maybe we can add s.con.Close() too

Varunram: oh, ok, I left it because it would eventually fall through and disconnect, but makes sense to disconnect here as well

uspv/msghandler.go
```diff
@@ -13,7 +13,11 @@ func (s *SPVCon) incomingMessageHandler() {
...
+			if s.randomNodesOK { // if user wants to connect to localhost, let him do so
+				s.Connect("yes") // really any YupString here
+			}
+			s.con.Close()
```
adiabat: this will connect then immediately close the connection if s.randomNodesOK is true

Varunram: noob mistake, sorry!

adiabat: No prob... still doesn't work though.  Now it quits the message handler after it re-connects.

PR #206, opened 2018-06-15
Vendored GitHub dependencies in prep for including (some of) them directly.

delbonis:
This might be more than we actually need.  Oops?


metalicjames:
ok to test

PR #208, opened 2018-06-15
Enforce minOutput rules in fund

metalicjames:
Don't let users fund channels that are impossible to close by using an initial send less than the minOutput value or where their own balance is similarly too low. 

In the future we might want to allow "top-heavy" channels that only have one output so that channels with 0 balance on one side can still be closed.


Varunram:
Finally a problem with the test itself lol

adiabat:
good catch, yeah we also need to test / make sure single output channels work well as funding with an intialSend of 0 could be pretty common

PR #209, opened 2018-06-15
Fixed Makefile being failing and gitignored the _workspace directory that the env.sh creates.

delbonis:


metalicjames:
add to whitelist

Makefile
```diff
@@ -6,6 +6,7 @@ GOBIN = $(shell pwd)/build/bin
 GO ?= latest
 
 lit:
-	build/env.sh go get . go build
```
Varunram: nice catch! I built upon this PR to add a bit more functionality.

PR #210, opened 2018-06-16
Update Makefile and README

Varunram:
Give the Makefile some more ammo to do more stuff and update the README to reflect changes in installing lit. Builds upon #209, so can be merged after /  in place of it.

Now, we can modify jenkins to just run `make test --with-python=true` (instead of the entire `go build`, `go test` and `python3` routine that we have right now)


delbonis:
May I suggest using a Jenkinsfile so we can update settings programmatically?

Varunram:
Oh, I guess I dunno much about `Jenkinsfile`s other than it does some cool pipelining stuff. Can it replace `make` in a simpler way? If so, would be cool..

delbonis:
No it's for setting up and invoking `make`, and then reporting build artifacts.  You mentioned we should configure Jenkins to do `make test --with-python=true`, but if we add that to the Jenkinsfile we woudn't have to have someone update the actual Jenkins to do it.

gertjaap:
I checked out the PR and then tried running `make`. I'm getting this error:

```
gertjaap@gjmacbook:~/go/src/github.com/mit-dci/lit$ make
build/env.sh go get ./...
# github.com/andlabs/ui
/usr/bin/x86_64-linux-gnu-ld: ../../andlabs/ui/libui_linux_amd64.a(libui-combined.o): relocation R_X86_64_32S against `.rodata' can not be used when making a PIE object; recompile with -fPIC
/usr/bin/x86_64-linux-gnu-ld: final link failed: Nonrepresentable section on output
collect2: error: ld returned 1 exit status
Makefile:12: recipe for target 'goget' failed
make: *** [goget] Error 2
```

Any reason for this? If so, should it be in the README?

Varunram:
No idea, but this seems to be orthogonal to the PR itself since the makefile does its job of `go get`ting github.com/andlabs/ui. Did you try searching for the error?

PR #214, opened 2018-06-18
Add Jenkinsfile

Varunram:
Added a jenkinsfile on @delbonis's suggestion - A separate PR because we may want more things to be added to this (and presumably if someone wants to build on this to have some nice features in the future, welcome!)

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,0 +1,14 @@
+#!groovy
+lit('lit') {
+currentBuild.result = 'SUCCESS'
+  try {
+  stage('Test') {
+
+    echo "Building lit"
+    sh 'make lit --with-python'
+   }
+ } catch(err){
+   currentBuild.result = "FAILURE"
+   throw err
+ }
+}


metalicjames:
How is this supposed to work?

delbonis:
We would need to modify the Jenkins configuration to "just run the jenkinsfile".

Varunram:
here you go: https://jenkins.io/doc/pipeline/tour/hello-world/

metalicjames:
Does this script adequately replace the setup we have already? It doesn't make sense to switch it over until there is "feature parity".

Varunram:
Afaik, this doesn't replace the setup in any way, this just shifts the setup of jenkins from the conf we have on the server side to this file.

delbonis:
You would need to tell Jenkins to use it as a "pipeline project", and then use the jenkinsfile from the repo.  And then all the important build work would be handled by the jenkinsfile.

Varunram:
`make lit --with-python` does what you just described. If you set GOPATH in advance ie. That could be done by Make too, so I don't see a reason why this can't replace the thing on Jenkins.

What I don't see however, is whether we need this, how maintainable it is, etc.

Varunram:
Closed in favour of #271

PR #215, opened 2018-06-18
Add docs script

Varunram:
Its *super* ugly right now because all the stuff is hardcoded. We must either standardise outputs in some format and parse it, but the problem here is that we'd have to add dlc commands to the help output of lit-af. `bitcoind` docs (the one at bitcoin-rpc.github.io earlier and at the bitcoin core webpage now) parses all the rpc help commands, so maybe we could have an option that does that. This would mean, however we repeat the help commands for that particular command. 

Live demo at https://varunram.com/litrpc/


gertjaap:
Hey, i like the idea of live docs like this, integrated into the codebase. I don't really follow though why it's coded in go, and generates documentation from hardcoded values. Could you elaborate your thinking behind this?

I could see how this is valuable conceptually, but then i'd expect the PR to at least contain an example of it being used from the lit-af code, so that we can see how the documentation ought to be centralized. This would also make it obvious that adding a new command requires you to fill these documentation structs. 

In the setup posed in this PR you can create a new command in lit-af without being aware that it's expected to be added into this docgen file also, so i'd prefer it to be a little more complete before we merge it in.

Thoughts?

Varunram:
. Hey, i like the idea of live docs like this, integrated into the codebase. I don't really follow though why it's coded in go, and generates documentation from hardcoded values. Could you elaborate your thinking behind this?

The reasoning behind the hardcoding is simple: The output of lit is not formatted and we don't have rpc output (help commands ie) for all the commands supported by lib. This is true for all commands in lit-af since we have to start the interface in order to see help.

> I could see how this is valuable conceptually, but then i'd expect the PR to at least contain an example of it being used from the lit-af code, so that we can see how the documentation ought to be centralized. This would also make it obvious that adding a new command requires you to fill these documentation structs.

Oh yeah, the "docs" are pretty much the help output from lit. I have yet to build upon them. But the idea of online docs is not to use lit-af and bitcoind does something similar as well.

> In the setup posed in this PR you can create a new command in lit-af without being aware that it's expected to be added into this docgen file also, so i'd prefer it to be a little more complete before we merge it in.

A new command in lit-af? not sure I get what you mean, you would ideally want a command that generates these docs locally? If so, would make sense as a standalone script or maybe in `lit` but I don't see how we'd use it in `lit-af`? Would be nice if you could expand upon this, thanks!

gertjaap:
So, I like the fact of having the documentation integrated in the code - but i don't see the added value of it being in the code if it's not an integral part of the code. Right now it's just a separate file that generates documentation from static code. Why wouldn't you just write this in a markdown file in stead? It seems overhead.

Unless, you integrated it in such a way that if you code something into lit-af you "automatically" generate the documentation for the command. Similar to how godoc generates documentation from the comments above functions.

So that we can run this on every build and update a documentation site. If you have to write hardcoded strings into a `.go` file that generates documentation when you run it, this seems like a weird set up to me.

Varunram:
>  Why wouldn't you just write this in a markdown file in stead? It seems overhead.
Well, you could, but remember markdown is overhead as well.

> Unless, you integrated it in such a way that if you code something into lit-af you "automatically" generate the documentation for the command. Similar to how godoc generates documentation from the comments above functions

This is an issue in itself I think, we must allow lit-af help subcommands to be called out of lit-af. Creating an issue for that.

Might be better to maintain this myself atm. Closing this.

PR #218, opened 2018-06-19
Vendored *just* btcd/btcutil libs.

delbonis:
I'll be pulling them out of dependencies and actually making them a part of Lit in another PR.  This is just to sync up with this history for now.


adiabat:
Cool - the areas where there's overlap are:
lnutil, coinparam
btcd, and btcutil.
Some of the workarounds for the btcsuite imports can be merged in to the modified library which could simplify things.

PR #220, opened 2018-06-19
Preliminary NAT traversal

Varunram:
Will improve upon this if merged, but need this so that we can work on top of this to implement more stuff.

### DIFF
--- a/litinit.go
+++ b/litinit.go
@@ -2,11 +2,14 @@ package main
 
 import (
 	"bufio"
+	"context"
 	"io"
 	"log"
+	"fmt"
 	"os"
 	"path/filepath"
 
+	UpnP "github.com/NebulousLabs/go-UpnP"
 	"github.com/jessevdk/go-flags"
 	"github.com/mit-dci/lit/lnutil"
 )
@@ -130,5 +133,32 @@ func litSetup(conf *config) *[32]byte {
 		log.Fatal(err)
 	}
 
+	// do UPnP port forwarding
+	// right now we fatal if we aren't able to port forward via upnp
+	// a question though is whether we should continue connceting without
+	// port forwardign
+	if conf.UPnP {
+		// Connect to router
+		var externalIPs []string
+		d, err := UpnP.DiscoverCtx(context.Background())
+		if err != nil {
+			fmt.Printf("Unable to discover router %v\n", err)
+			log.Fatal(err)
+		}
+		// Get external IP
+		ip, err := d.ExternalIP()
+		if err != nil {
+			fmt.Printf("Unable to get external ip %v\n", err)
+			log.Fatal(err)
+		}
+		log.Printf("Your external IP is %s", ip)
+		// Forward peer port
+		err = d.Forward(uint16(conf.Rpcport), "lnd peer port")
+		if err != nil {
+			fmt.Printf("UpnP: Unable to forward pear port ip %v\n", err)
+			log.Fatal(err)
+		}
+		externalIPs = append(externalIPs, ip)
+	}
 	return key
 }
--- a/litinit.go
+++ b/litinit.go
@@ -2,16 +2,16 @@ package main
 
 import (
 	"bufio"
-	"context"
+	"fmt"
 	"io"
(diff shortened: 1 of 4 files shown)


litinit.go
```diff
@@ -130,5 +133,32 @@ func litSetup(conf *config) *[32]byte {
...
+		}
+		// Get external IP
+		ip, err := d.ExternalIP()
+		if err != nil {
+			fmt.Printf("Unable to get external ip %v\n", err)
+			log.Fatal(err)
+		}
+		log.Printf("Your external IP is %s", ip)
+		// Forward peer port
+		err = d.Forward(uint16(conf.Rpcport), "lnd peer port")
```
gertjaap: We shouldn't NAT the RPC Port, but the lndc port (2448). That's not configurable rn, because you pass it to `listen` (or configure the automatic Listen port that is used to trigger `listen` on startup). 

It would make sense to do the NAT punch when `listen` is executed, because we know the port we need to punch at that time only.

I noticed when i tried running this, and found the uPnP was opening port 8001 on my router.

This does make me feel this wasn't tested for actual functioning.

Varunram:
thanks, I tested it on 2448, but replaced with the wrong port before PRing. Will udpate.

Varunram:
Regarding the port, this is initialised via `lit` and not `lit-af` so we'd need to pass the config in `lit-af`. I did this in another PR while implementing Tor support, so will link your comment over there and hardcode the listening port here for now.

gertjaap:
I don't see the reason for adding a *manual* command for NAT. This is not an improvement for the user from your last commit. What would've been wrong with just fixing the problem and running the NAT punch when the user executes (or lit automatically executes) Listen? We know the port that needs punching right there.

Varunram:
Anything that is done automatically has to be thought over carefully - what if NAT punching is banned for some strange reason in the jurisdiction of the user? I don't think we should assume what the user wants hence I see no reason why it should be enabled by default.

gertjaap:
It isn't enabled by default! You made a boolean switch for upnp in the config which is off by default. We should (obviously) only run the code if this is switched on - but i thought that was implicit from the previous commit(s).

All I mean by my comment is, if uPnP is enabled, we should automatically trigger it at the right point in time - which is: when we start listening, we know the port we're listening on, we should punch it through on the firewall.

Varunram:
Oh, I see what you mean ; my mistake, there's no reason why this command should be in lit and in `lit-af`, removing it from `lit`. The user can trigger this from lit-af if he wants to (I thought we were debating on this part, sorry!)

gertjaap:
I don't want to be a nag, but i'm gonna anyway:

I would remove it from `lit-af` and keep the upnp switch in LIT. NAT Punchthrough should not be controlled from the UI. By keeping it in `lit-af` it means you will also have to add it to the webui separately, and you're then implementing it in two places.

I sincerely think it's best to trigger NAT punch when you start to listen, based on the uPnP setting being enabled. So from the host process and not from the client.

gertjaap:
We added the option to automatically listen for connections on startup. That will not / can not trigger NAT in your current setup, because you chose to keep NAT a separate RPC call that you trigger from the client.

So also again, if we build a Web UI, you have to build the same logic (allow picking nat yes/no, calling two functions in stead of one) into that client as well.

This code really needs to be in the qln.TcpListener function, and be used if a global configuration variable in `lit.conf` or the command line is switched to true/yes.

Varunram:
I think the NAT option is something similar to `lis` or `con`, which really must be triggered by the UI because these are fairly advanced options which need not be done at the daemon level.  Also, we need to call `lis` with `nat` inside it, which is the same as calling `nat` and `lis` separately, so I don't think we have much of a saving there. Might be better to loop @adiabat in to hear his views on this, so that we have a majority of opinion on some side.

Varunram:
Since bitcoin core does this via a bool flag, we'll be doing something similar.

gertjaap:
Discussed during the meeting. Based on the setting in lit.conf (upnp enabled), we punch NAT when listening starts.

PR #221, opened 2018-06-19
Pulled btcd/btcutil dependencies into project, removing vendor directory.

delbonis:
I hope GitHub doesn't make this look like I personally wrote all this code that's I'm moving around.


delbonis:
Closing in favor of slightly cleaner method.

PR #224, opened 2018-06-20
format output, make things look nicer/colorful

Varunram:
Mostly trivial changes

### DIFF
--- a/cmd/lit-af/shell.go
+++ b/cmd/lit-af/shell.go
@@ -3,7 +3,9 @@ package main
 import (
 	"bytes"
 	"fmt"
+	"sort"
 	"strconv"
+	"strings"
 
 	"io/ioutil"
 	"net/http"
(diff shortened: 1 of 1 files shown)


cmd/lit-af/shell.go
```diff
@@ -223,14 +225,42 @@ func (lc *litAfClient) Ls(textArgs []string) error {
 		return err
 	}
 	if len(cReply.Channels) > 0 {
-		fmt.Fprintf(color.Output, "\t%s\n", lnutil.Header("Channels:"))
+		fmt.Fprintf(color.Output, "\t%s\n", lnutil.Header("Closed Channels:"))
```
gertjaap: I think this should just read `Channels` and not `Closed Channels`

cmd/lit-af/shell.go
```diff
@@ -223,14 +225,42 @@ func (lc *litAfClient) Ls(textArgs []string) error {
...
 	for _, c := range cReply.Channels {
 		if c.Closed {
-			fmt.Fprintf(color.Output, lnutil.Red("Closed  "))
+			closedChannels = append(closedChannels, c)
+		} else {
+			openChannels = append(openChannels, c)
+		}
+	}
+
+	for _, c := range closedChannels {
```
gertjaap: Maybe swap the channel sets and put the closed ones at the bottom? Makes sense to put the active ones on top as it sort of naturally feels that whatever is on top is more important.

cmd/lit-af/shell.go
```diff
@@ -331,7 +361,7 @@ func printHelp(commands []*Command) {
 func printCointypes() {
 	for k, v := range coinparam.RegisteredNets {
 		fmt.Fprintf(color.Output, "CoinType: %s\n", strconv.Itoa(int(k)))
-		fmt.Fprintf(color.Output, "└────── Name: %s,\tBech32Prefix: %s\n", v.Name, v.Bech32Prefix)
+		fmt.Fprintf(color.Output, "└────── Name: %s,%sBech32Prefix: %s\n\n", v.Name, strings.Repeat(" ", 13-len(v.Name)), v.Bech32Prefix)
```
gertjaap: Alternatively, i think you can use `%-13s` in the format string and append the comma to v.Name in the parameters:

```
fmt.Fprintf(color.Output, "└────── Name: %-13sBech32Prefix: %s\n\n", v.Name + ",", v.Bech32Prefix)
```

cmd/lit-af/shell.go
Varunram: Will do

cmd/lit-af/shell.go
Varunram: Thx, this looks a bit ugly, yeah

cmd/lit-af/shell.go
Varunram: Will change along with other updates.

cmd/lit-af/shell.go
Varunram: Okay, just remembered something, why I put it on top was because people would have to scroll up and see the open channels otherwise.

gertjaap: Okay, that stems from the fact that `ls` outputs so much at once. Created an issue #230 to try and fix that. I think we can resolve this remark that way too, and for this PR the chosen sequence makes sense.

PR #226, opened 2018-06-21
Make litecoin regtest work

Varunram:
litecoin regtest seems to behave weirdly, get headers and inHeaders return different blocks making attachHeight -1.


gertjaap:
I've checked the code, as @Varunram mentioned it's a little "Hacky". The reason it doesn't work is twofold: the genesis hash for litecoin regtest mismatches, and uspv does not accept nodes that identify as "LitecoinCore". I fixed that in #226

PR #227, opened 2018-06-22
BOLT8 conformance rebase

Varunram:
Builds upon #164, rebased on top of <a href="https://github.com/mit-dci/lit/commit/c4ded8d53585e1b7474b40b9f71cbe09d029d22e">master</a> and defaulted to lit config. Added LICENSE to avoid conflicts and a timeout to tracker lookup since that conflicted with the RPC call timeout that we have in case the ip lookup fails. Attributed @yaslama for the relevant commits. Should be good to merge.


Varunram:
import issues, fixing.

Varunram:
retest this please

Varunram:
retest this please

Varunram:
Looks like the timeout of 10s  was too small for `lit-af`.

gertjaap:
Commit seems to be in order, tested the functionality and that works. But I don't like the fact that we're ditching the `ln1...` addresses in favor of full public keys. We should look at that before merging this in.

Varunram:
@adiabat Requesting review, this has been around for a very long time and it'd be nice if we can decide  whether we're going to adopt this or not to avoid conflicts like that in #262.

Varunram:
Looks like the test failed due to import stuff, pushing again..

Varunram:
No reason not to use bech32. Closing.

PR #228, opened 2018-06-25
Expose LIT address into peer list

adamcollier1:
To identify a channel to a peer by lit address

### DIFF
--- a/qln/netio.go
+++ b/qln/netio.go
@@ -179,6 +179,7 @@ func (nd *LitNode) OutMessager() {
 type PeerInfo struct {
 	PeerNumber uint32
 	RemoteHost string
+	LitAdr 	   string
 	Nickname   string
 }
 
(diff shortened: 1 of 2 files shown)


metalicjames:
ok to test

qln/netio.go
```diff
@@ -179,16 +179,19 @@ func (nd *LitNode) OutMessager() {
...
 	RemoteHost string
+	LitAdr 	   string
 	Nickname   string
 }
 
 func (nd *LitNode) GetConnectedPeerList() []PeerInfo {
 	var peers []PeerInfo
 	for k, v := range nd.RemoteCons {
 		var newPeer PeerInfo
+		var pubArr,_ = nd.GetPubHostFromPeerIdx(k)
```
gertjaap: Since GetPubHostFromPeerIdx reads data from the database, it seems to make sense to swap this out for:

```go
var pubArr [33]byte
copy(pubArr[:], v.Con.RemotePub.SerializeCompressed())
```
Since the remote pubkey is already in memory.

gertjaap: Since GetPubHostFromPeerIdx reads data from the database, it seems to make sense to swap this out for:

```go
var pubArr [33]byte
copy(pubArr[:], v.Con.RemotePub.SerializeCompressed())
```
Since the remote pubkey is already in memory.

adiabat:
Looks good, thanks!

PR #232, opened 2018-06-27
Fixes Litecoin Regtest

gertjaap:
The problem with Litecoin regtest was the fact that the genesis hash mismatched in our code, and that nodes identifying as `LitecoinCore` were denied. This PR fixes that issue.


uspv/init.go
```diff
@@ -143,7 +143,7 @@ func (s *SPVCon) Handshake(listOfNodes []string) error {
 		return fmt.Errorf("Remote node version: %x too old, disconnecting.", mv.ProtocolVersion)
 	}
 
-	if !(strings.Contains(mv.UserAgent, "Satoshi") || strings.Contains(mv.UserAgent, "btcd")) && (len(listOfNodes) != 0) {
+	if !((strings.Contains(s.Param.Name, "lite") && strings.Contains(mv.UserAgent, "LitecoinCore")) || strings.Contains(mv.UserAgent, "Satoshi") || strings.Contains(mv.UserAgent, "btcd")) && (len(listOfNodes) != 0) {
```
Varunram: Hopefully we can get something better as we move on adding other coins instead of a whitelist, will test VTC testnet and update this hopefully.

btcutil/btcd/chaincfg/genesis.go
```diff
@@ -175,10 +175,10 @@ var liteCoinTestNet4GenesisBlock = wire.MsgBlock{
 
 // liteCoinRegTestGenesisHash is the first hash in litecoin regtest
 var liteCoinRegTestGenesisHash = chainhash.Hash([chainhash.HashSize]byte{ // Make go vet happy.
-	0x06, 0x22, 0x6e, 0x46, 0x11, 0x1a, 0x0b, 0x59,
-	0xca, 0xaf, 0x12, 0x60, 0x43, 0xeb, 0x5b, 0xbf,
-	0x28, 0xc3, 0x4f, 0x3a, 0x5e, 0x33, 0x2a, 0x1f,
-	0xc7, 0xb2, 0xb7, 0x3c, 0xf1, 0x88, 0x91, 0x0f,
+	0xf9, 0x16, 0xc4, 0x56, 0xfc, 0x51, 0xdf, 0x62,
```
Varunram: I guess this being hardcoded in btcd is why the error came in the first place

PR #233, opened 2018-06-27
Cleaned up new btcutil stuff, removing some redundancies

delbonis:
This fully splits apart the btcd stuff, getting rid of everything I could tell that could be removed.  Also I updated the README a bit, adding a bit more documentation.


delbonis:
There's a lot of refactoring that I'm thinking about how to actually go about that would reduce the need for scarier rebases later on by slimming down the dependency graph between components in Lit.  Gonna wait until after we have HTLCs merged.

delbonis:
Oh dear this is quite a messy merge.

delbonis:
Ok that should be good now.  Not sure why it says "2 days ago".

gertjaap:
Checked out the code and basic functionality is in tact.

PR #241, opened 2018-06-28
Fix the problem when short signature

tnakagawa:
Signature of the special transaction is short size.

I fixed that problem , and added test case.

Signature before fixed :
3041021e0a15f55034e27e85003ca2e970d2325162592186975d29fc49b4ee24eb36021f4944c8cf4db9765d8226686ae7e9407e7a384d58b50ca786b40ddad23a6b8601

Error message before fixed :
Can't compress; sig length is 67


Varunram:
ok to test

adiabat:
Thanks -- I'm surprised because this seems like it must occur (once every 2^32 signatures?) already.
even 65 technically could fail I think, but should (almost) never happen.
ありがとう御座います

Issue #240, opened 2018-06-28
Re-orging is still broken - breaks testnet sync


metalicjames:
```
2018/06/27 21:11:55.777661 asking for headers due to inv block
2018/06/27 21:11:55.777695 got header tip height 1327070
2018/06/27 21:11:55.778185 get headers message has 18 header hashes, first one is 0000000029caccca59f6ea2e2e87870fd4c5f42e77cd7237071c30d7112b03ac
2018/06/27 21:11:55.891456 got 3 headers. Range:
00000000264c664adf909677141b6a039f004c02c6047becbccb7b97b3a72325 - 000000002128c5b8fbe782216cca37e3cb8de8ff5278e703cfe801af397461a0
2018/06/27 21:11:55.891481 header file position: 5688160
2018/06/27 21:11:55.891494 made 2017 header slice
2018/06/27 21:11:55.900810 Header 00000000264c664adf909677141b6a039f004c02c6047becbccb7b97b3a72325 attaches at height 1327068
2018/06/27 21:11:55.900817 reorg from height 1327070 to 1327071
2018/06/27 21:11:55.900903 Added 3 headers OK.
2018/06/27 21:11:55.900914 got header tip height 1327071
2018/06/27 21:11:55.900973 HeightHandler: oh no, reorg!
2018/06/27 21:11:55.901038 Rollback height 1327068
2018/06/27 21:11:55.901059 get headers message has 18 header hashes, first one is 000000002128c5b8fbe782216cca37e3cb8de8ff5278e703cfe801af397461a0
2018/06/27 21:11:55.901072 tx height 1326447
2018/06/27 21:11:55.901090 tx height 1326447
2018/06/27 21:11:55.901105 tx height 1326447
2018/06/27 21:11:55.901121 tx height 1326449
2018/06/27 21:11:55.901137 tx height 1326526
2018/06/27 21:11:55.901154 tx height 0
2018/06/27 21:11:55.901170 tx height 1326450
2018/06/27 21:11:55.901186 tx height 1326450
2018/06/27 21:11:55.901202 tx height 0
2018/06/27 21:11:55.901216 tx height 1326449
2018/06/27 21:11:55.901228 Rollback db.  0 utxos lost
2018/06/27 21:11:56.031270 got inv.  Contains:
2018/06/27 21:11:56.031317 	0)MSG_BLOCK : 000000007cb9d64cf2dd08ba972e178a6890b55b7855cc0ee04aa3ab4d886655
[...]

PR #248, opened 2018-06-28
Bugfixes for testnet

Varunram:
Dodges the zero header bit bug (no problem in reorg code / our side, the incoming headers themselves have this set to 0, dunno why) and fixed a bug where you can't join the chain after a deep re-org. Also removed some noob error statements.


gertjaap:
This is for Issue #240

metalicjames:
Possible fix for #240

Issue #231, opened 2018-06-27
Problem with CheckHelpCommand


gertjaap:
In lit-af, we added a `CheckHelpCommand` utility function to check the required number of parameters and print out the command format/description when you specify `-h`. However, in this new setup the control flow of the main function is not ended when someone enters `-h` because it will only break on an error. Consequence is that command that only require 1 parameter will continue to execute trying to parse `-h` as the first parameter.

Most obvious example is when you execute `close -h`:

```
lit-af# close -h
entered command: close -h
close <channel idx>
Cooperatively close the channel with the given index by asking
the other party to finalize the channel pay-out.
See also: break
close error: strconv.Atoi: parsing "-h": invalid syntax
```

Varunram:
Nice catch, I'll take this one up.

Varunram:
Fixed in #250

PR #250, opened 2018-06-29
Add stopEx bool to CheckHelpCommand

Varunram:
Earlier, we didn't have a way whether we should stop execution, instead we relied on the code that follow to do that for us. But this caused side effects like that in #231, which should now be solved with the introduction of this bool.

### DIFF
--- a/cmd/lit-af/walletcmds.go
+++ b/cmd/lit-af/walletcmds.go
@@ -42,8 +42,8 @@ var sweepCommand = &Command{
 
 // Send sends coins somewhere
 func (lc *litAfClient) Send(textArgs []string) error {
-	err := CheckHelpCommand(sendCommand, textArgs, 2)
-	if err != nil {
+	err, stopEx := CheckHelpCommand(sendCommand, textArgs, 2)
+	if err != nil || stopEx {
 		return err
 	}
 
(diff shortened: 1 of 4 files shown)


adiabat:
Can you change it from error, bool to bool, error?  Other functions we use have errors at the end.
Also, can you document in a function definition comment what the bool means.  thanks

cmd/lit-af/walletcmds.go
```diff
@@ -80,8 +80,8 @@ func (lc *litAfClient) Send(textArgs []string) error {
 
 // Sweep moves utxos with many 1-in-1-out txs
 func (lc *litAfClient) Sweep(textArgs []string) error {
-	err := CheckHelpCommand(sweepCommand, textArgs, 2)
-	if err != nil {
+	err, stopEx := CheckHelpCommand(sweepCommand, textArgs, 2)
+	if err != nil || stopEx {
```
adiabat: This will return a nil error if stopEx is true.  Is this the functionality we want?  Seems strange to break execution with no error.

Varunram: Well, stopEx checks whether the user has requested the help command. Since the user hasn't input something wrong, we don't return an error. Returning one will bring us back to the present case that we have :)

Varunram:
retest this please

cmd/lit-af/walletcmds.go
gertjaap: In the old situation we would also return nil if the `-h` parameter was passed:

(one instance in an older version of this code:)

```
	if len(textArgs) > 0 && textArgs[0] == "-h" {
		fmt.Fprintf(color.Output, historyCommand.Format)
		fmt.Fprintf(color.Output, historyCommand.Description)
		return nil
	}
```

adiabat: OK that makes sense.  Not actually an "error"

adiabat:
Looks good, thanks!

PR #252, opened 2018-06-30
move webui inside lit

Varunram:
added an electron build script under make. Also added modified package.json and added an electron script inside webui.


Varunram:
Hm, I think I published an old commit here again, will push

Varunram:
Okay, this works, but I can't certify the degree of bugginess due to the different envs. Will report back tomorrow.

Varunram:
Okay, I think I get why we get a dark screen if it doesn't work. ITs because `npm run build` fails sometimes. I don't know how to fix this, so opening a issue in the main `webui` repo to see if we can get a fix there.

Varunram:
Short note on what this does: It builds an electron app out of the code sitting in `webui` right now and then builds it into two things - first is a native .exe, .dmg, etc and the second is your normal app over at `dist/`. Looks like the build didn't' work due to a small problem with `package.json` but thanks to Joe, everything works fine now.

gertjaap:
This seems to work when tested. I had to install `yarn` (i think the dependencies should be updated in the main LIT readme for that? Also mention of the webui in there makes sense, since it's part of the main makefile).

Once this is merged, i think we should look at building the webui and serving it from the lit daemon too (to not have to run a separate executable for the webui, just use a browser you already have to http://localhost:<port>/ )

adiabat:
Merging in; a longer term solution will have the lit binary serving the UI, as @gertjaap said.  
This is... non-trivial!  Ideally people could build lit AND get the UI with just go-get.

PR #253, opened 2018-07-01
Connect lit via tor

Varunram:
Connect lit via tor. Lots of things done in this PR, built on top of brontide. 

When starting lit, do `./lit --tor.active --tor.v2 --tn3 <host> -v`

To connect to some host, run `con adr@list.onion tor true` from `lit-af` and then you should be connected to a remote onion address. Connecting to a clearnet address via tor is possible as well, normal syntax should work.  Connecting via tor to remote Hosts for sync and connecting to a peer via tor is not possible (no bad opsec, sorry).

NOTE: This PR will fail tests because the tests would have to be rewritten as well, will rewrite. This also moves all config stuff to config/ because we need to access user provided config from lit-af and uspv as well.

Will fix tests towards the end after all incorporating all suggestions, doesn't make sense to write tests for constantly changing code.


Varunram:
Since this is built on brontide, will merge after the BOLT PR is merged.

gertjaap:
Since Brontide is now merged (right?) , could we finish this?

Varunram:
Will update, thanks!

delbonis:
Can I suggest using `--tor-active` and `--tor-v2` instead of the `.` notation?  it looks a little cleaner.

PR #255, opened 2018-07-04
Detect whether coin daemon is running before funding

Varunram:


wallit/init.go
```diff
@@ -107,8 +109,27 @@ func NewWallit(
 
 	// deal with incoming height
 	go w.HeightHandler(incomingBlockheight)
-
-	return &w
+	if !hookFail {
+		switch p.Name {
```
adiabat: I think you can do something like 
```
return &w, p.HDCoinType, nil
```
instead

Varunram: Yeah, was about to get to this, got stuck at the jenkins file thing. updating, thanks!

metalicjames:
retest this please

Varunram:
Rebased

PR #257, opened 2018-07-05
Added RPC API docs.

delbonis:
No code changes made, although I'm looking for comments on the doc format itself.


adiabat:
Looks good to me.

Further docs we can add would include something about possible error messages and why / when they're returned.  In depth description of errors (more so than the error text itself) can help when people can't get the API to work.

PR #261, opened 2018-07-05
Don't do anything if we don't enable --tower

Varunram:
Earlier nodes could remotely shut down a non tower peer peer by sending them tower messages. Now, we just ignore those messages. We could store it in some local db and give it to them for niceness (since they don't know we aren't a watchtower and we don't / can't tell them right now).


adiabat:
cool... would also be good to get the watchtower functionality working again & robust

PR #262, opened 2018-07-06
Allows connecting to hostnames (again)

gertjaap:
Somehow when implementing the ipv4 / ipv6 support, the functionality to be able to connect to hostnames was lost. I used this in my dockerized test environments (connecting to `{adr}@lit01` for instance)

Previously, i was able to connect to`{lnaddr}@{hostname}[:port]`. But this no longer works since hostnames are not recognized as valid IPs. Which you can turn them into by resolving them.

Added code to resolve hostnames.

Was unable to test this with IPv6 (as I don't have a DNS returning IPv6 addresses). If anyone knows how I can test that, i'd be happy to hear. For IPv4 this works.

### DIFF
--- a/lndc/conn.go
+++ b/lndc/conn.go
@@ -13,9 +13,9 @@ import (
 
 	"golang.org/x/net/proxy"
 
+	"github.com/codahale/chacha20poly1305"
 	"github.com/mit-dci/lit/btcutil/btcec"
 	"github.com/mit-dci/lit/crypto/fastsha256"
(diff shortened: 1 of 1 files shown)


delbonis:
I think that it's already been decided that we're not bothering with BOLT8 anytime soon due to the fundamental design differences.

Varunram:
There are no fundamental differences, just a small problem with `ln` addresses, which are overshadowed by the improved feature set of the new library.

PR #263, opened 2018-07-06
Add noise_XX protocol to lndc

Varunram:
Please merge this or #227, depending whether we want / don't want hex addresses. This also increases the length of the bech32 addresses, since the ones we used didn't actually have checksums.

EDIT: bech32 ftw


Varunram:
Looks like noise_test failed, changing that.

gertjaap:
As discussed during the meeting - we need to adjust this PR to support the shorter bech32 format like we use it right now. It needs an extra exchange at the start of the connection to establish the keys.

Varunram:
Need an extra message to look up the pubkey hash and get their pubkey.

delbonis:
Might want to update the README in there and then copy the license from the lnd project.

adiabat:
The modifications we'd need to support shorter addresses don't look too complex.

Basically we need an act 0 in cases where the dialer does not know the full public key of the listener.  

Instead, the dialer sends a "SendPKH" message where it sends the pubkey hash that it knows.  (Or maybe send the hash of that... not sure if that helps but that's also possible)

The listener then checks if the PKH it just received matches the pubkey it's using (or one of the pubkeys it's using).  The listener then responds with "RespondPK", which contains the full pubkey.  

The dialer receives the RespondPK message, checks that the pubkey fits the PKH it knows about, and errors out if it doesn't.  

From there, the act1 / act2 can proceed as in the code here.  Some flag needs to be set in the dialer's initial message to indicate whether it's sending over a PKH or act1.

adiabat:
Having an initial challenge / response with the PKH / PK loses the identity hiding for the listener, which is really not great. 

Looking at the noise protocol, we need XX instead of XK.  http://noiseprotocol.org/noise.html#interactive-patterns

This would be a substantial modification, but I think it's worth it, as it allows authentication when both parties only know each other's pubkey hashes, as well as trust-on-first-use without changing the underlying messaging.

adiabat:
We can try using this package:
https://github.com/flynn/noise but need to modify this to allow secp256k1 points  instead of just curve25519.  Not sure if modifying the brontide code or the flynn/noise code will be easier

Varunram:
Updated to work with `Noise_XX`.

Varunram:
Looks like two nodes can't start on localhost at the same time. Will investigate and fix.

Varunram:
retest this please

Varunram:
retest this please

Varunram:
restest this please

Varunram:
retest this please

lndc/README.md
```diff
@@ -0,0 +1,19 @@
...
+The lndc package implements a secure crypto messaging protocol based off of
+the [Noise Protocol Framework](http://noiseprotocol.org/noise.html). The
+package exposes the raw state machine that handles the handshake and subsequent
+message encryption/decryption scheme. Additionally, the package exposes a
+[net.Conn](https://golang.org/pkg/net/#Conn) and a
+[net.Listener](https://golang.org/pkg/net/#Listener) interface implementation
+which allows the encrypted transport to be seamlessly integrated into a
+codebase.
+
+The secure messaging scheme implemented within this package uses `NOISE_XX` as the handshake for authenticated key exchange.
```
adiabat: Should mention here that this is not what the brontide code does and it isn't compatible

Varunram: Thanks, will do.

litrpc/netcmds.go
```diff
@@ -49,7 +49,6 @@ func (r *LitRPC) Connect(args ConnectArgs, reply *StatusReply) error {
 	if err == nil {
 		// get peer from address book
 		pubArr, host := r.Node.GetPubHostFromPeerIdx(uint32(peerIdxint))
-
```
adiabat: please remove

Varunram: you mean the empty line?

lndc/netio.go
```diff
@@ -1,93 +0,0 @@
...
-	"bytes"
-	"encoding/binary"
-	"fmt"
-	"io"
-	"net"
-)
-
-// maxMsgSize is the longest message supported.  This is payload size, not
-// encrypted size.
-const maxMsgSize = 1 << 24
```
adiabat: We need to keep this; DLC message are using it.
(though I suppose it could be (1<<24)-1 instead)

PR #266, opened 2018-07-09
Clarified control flow and cleaned up warnings in uspv connection control code

delbonis:


Varunram:
Node is closed for some weird reason, we should check regarding that.

uspv/header.go
```diff
@@ -26,7 +26,7 @@ func min(a, b int) int {
 	return b
 }
 func moreWork(a, b []*wire.BlockHeader, p *coinparam.Params) bool {
-	var isMoreWork bool = false
+	var isMoreWork = false
```
Varunram: This could be better off as `isMoreWork := false` I guess

uspv/init.go
```diff
@@ -73,32 +74,32 @@ func (s *SPVCon) GetListOfNodes() ([]string, error) {
 }
 
 // DialNode receives a list of node ips and then tries to connect to them one by one.
-func (s *SPVCon) DialNode(listOfNodes []string) error {
```
Varunram: We _might_ use DialNode in other places in the future (see: Tor routing PR), so lets leave this exported :)

uspv/init.go
```diff
@@ -73,32 +74,32 @@ func (s *SPVCon) GetListOfNodes() ([]string, error) {
...
+func (s *SPVCon) dialNode(listOfNodes []string) (net.Conn, error) {
 	// now have some IPs, go through and try to connect to one.
-	var err error
-	for i, ip := range listOfNodes {
+	for _, ip := range listOfNodes {
 		// try to connect to all nodes in this range
-		var conString, conMode string
+		//var conString, conMode string
```
Varunram: lets remove this if its not being used.

uspv/init.go
```diff
@@ -217,6 +219,7 @@ func (s *SPVCon) Connect(remoteNode string) error {
 			continue
 		}
 	}
+	s.con = con
```
Varunram: we can directly write to `s.con`?

uspv/init.go
```diff
@@ -273,9 +276,9 @@ func (s *SPVCon) openHeaderFile(hfn string) error {
 			// if StartHeader is defined, start with hardcoded height
 			if s.Param.StartHeight != 0 {
 				hdr := s.Param.StartHeader
-				_, err := b.Write(hdr[:])
-				if err != nil {
-					return err
+				_, err2 := b.Write(hdr[:])
```
Varunram: Do we need `err` in this scope? else, we're better off with `err` :)

delbonis:
retest this please

Varunram:
Merge conflict again :)

Varunram:
Also, looks like you have two test hooks connected to this PR - any reasons why?

delbonis:
I don't know why there's two of them.

delbonis:
retest this please

delbonis:
retest this please

delbonis:
Bah why is this build failing?  I didn't change anything related to what the error would be about.

uspv/header.go
```diff
@@ -26,13 +26,13 @@ func min(a, b int) int {
 	return b
 }
 func moreWork(a, b []*wire.BlockHeader, p *coinparam.Params) bool {
-	var isMoreWork bool = false
```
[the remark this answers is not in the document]
delbonis: It's correct but a bit redundant, `go vet` just complains.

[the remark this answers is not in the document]
Varunram: I see, weird..

delbonis:
Ok this branch keeps getting harder and harder to rebase every time I need to and the history is getting really freaky at this point, but it should actually work now finally, can someone merge it now?

uspv/init.go
```diff
@@ -210,8 +203,10 @@ func (s *SPVCon) Connect(remoteNode string) error {
 	}
 	handShakeFailed := false // need to be in this scope to access it here
 	connEstablished := false
+	var con net.Conn
 	for len(listOfNodes) != 0 {
-		s.con, err = s.DialNode(listOfNodes)
+		con, err = s.DialNode(listOfNodes)
+		s.con = con
```
Varunram: Why would we return `net.Conn` from `DialNode` when we assign the `con` returned to `s.con` in the next line? It would make sense if we used `con` somewhere else, but we don't, so I think its better we assign a value to `s.con` directly inside `DialNode`

uspv/init.go
```diff
@@ -104,20 +103,14 @@ func (s *SPVCon) DialNode(listOfNodes []string) (net.Conn, error) {
 		} else {
 			con, err = net.Dial(conMode, conString)
 		}
-
-		if err != nil {
-			if i != len(listOfNodes)-1 {
-				log.Println(err.Error())
-				continue
-			} else if i == len(listOfNodes)-1 {
```
Varunram: we do need to catch this error or else we'll loop infinitely in case we can't connect :)

test/litnode.py
```diff
@@ -41,6 +42,7 @@ def start_node(self):
         logger.debug("Starting litnode %d with args %s" % (self.index, str(self.args)))
         assert os.path.isfile(LIT_BIN), "lit binary not found at %s" % LIT_BIN
         self.process = subprocess.Popen([LIT_BIN] + self.args, stdout=subprocess.DEVNULL, stderr=subprocess.DEVNULL)
+        time.sleep(5)
```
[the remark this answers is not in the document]
Varunram: Also ping on this

uspv/init.go
Varunram: needs change before merge :)

Issue #235, opened 2018-06-27
Introduce the option for remote control by a delegated client


gertjaap:
General idea: Generate a keypair remotely, and authorize the public key in LIT for certain commands (either view-only or also payment-related) and optionally limited to a certain amount. 

Then LIT can accept signed RPC requests from a remote peer wrapped in normal lndc messages. As long as the public key matches an authorized one, the signature on the request is valid, and the authorization checks out, LIT will execute a the command (return balance info, execute payment, etc).

Usage scenarios:

- Cell phone app - It generates a key pair on the phone, and you can authorize the public key in LIT. The phone can scan a LIT address / amount from a QR-code, sign the payment request, and send it to your “master” LIT node that will execute the payment
- Card terminal - There’s smartcards (contact and contactless) that do on-card signing. It could allow for a payment terminal to make a signing request, you sign it with the smart card, transfer the signed request to your LIT node that executes the payment

gertjaap:
Updated PR #312

PR #267, opened 2018-07-10
[WIP] Remote control

gertjaap:
This implements #235 . The idea is as follows:

1. You generate a EC Keypair
2. You then authorize the public key in the lit node using the `rcauth {pubkey} true` command
3. You can then sign a JSON request with your key. JSON is formed like a RPC call:

```
{
  "method" : "Say",
  "args" : {
    "Peer": 1,
    "Message": "hello"
  }
}
```

4. Any LNDC node can deliver a message to the lit node (message type `0xB0`) that contains:

- PubKey
- JSON
- Signature

5. The lit node verifies if the pubkey is authorized
6. The lit node verifies the signature
7. The lit node carries out the command in the JSON by sending it to its RPC server
8. The lit node returns the response in a message type `0xB1`

I've tested this with the `Say` command so far, which works. Tried with invalid pubkeys and signatures the requests get rejected properly

Next steps:

- [ ] Add IDs to match request/response
- [ ] Allow more finegrained authorization than just "true / false"
- [ ] Add nonces to the requests and disallow double nonces to prevent replays


Varunram:
retest this please

cmd/lit-af/netcmds.go
```diff
@@ -164,3 +180,50 @@ func (lc *litAfClient) Say(textArgs []string) error {
...
+
+func (lc *litAfClient) RemoteControlAuth(textArgs []string) error {
+	stopEx, err := CheckHelpCommand(rcAuthCommand, textArgs, 2)
+	if err != nil || stopEx {
+		return err
+	}
+	args := new(litrpc.RCAuthArgs)
+	reply := new(litrpc.StatusReply)
+
+	pubKey, err := hex.DecodeString(textArgs[0])
```
Varunram: hey, just trying this out, why do we use a hex encoded argument here? remnants of the old brontide stuff?

lit.go
```diff
@@ -174,6 +178,16 @@ func main() {
 		node.AutoReconnect(conf.AutoListenPort, conf.AutoReconnectInterval)
 	}
 
+	origin := "http://127.0.0.1/"
```
Varunram: I assume this is for testing direct conns to lit-af?

Varunram: I assume this is for testing direct conns to lit-af?

cmd/lit-af/netcmds.go
Varunram: hey, just trying this out, why do we use a hex encoded argument here? remnants of the old brontide stuff?

lndc/conn.go
```diff
@@ -72,9 +72,24 @@ func parseAdr(netAddress string) (string, string, error) {
 	} else if colonCount >= 5 {
 		conMode = "tcp6"
 		return netAddress, conMode, nil
-	} else {
-		return "", "", fmt.Errorf("Invalid ip")
+	} else if colonCount == 1 {
```
Varunram: Nice, I wish we could remove the ugly method of checking ips though, there must be a better way..

Varunram:
retest this please

gertjaap:
I'm gonna redo this.

Issue #230, opened 2018-06-27
Split the ls command into multiple ones, clean up the main command


gertjaap:
OK, so while reviewing #224 I came to the idea of splitting up the `ls` command. 

When you issue `ls` without any parameters, it should stick to:

- Showing only open and possible pending channels
- Showing only the latest 2-3 addresses
- Show your listening address (if listening)

Then, we can add subcommands to `ls` that give more info, like

- `ls channels` would show all channels including closed ones
- `ls utxo` would show all UTXOs
- etcetera

Thoughts?

Varunram:
Sounds good to me, I've always not liked the idea of the console getting flooded with lots of output.

Varunram:
Maybe once we have this in place, we could look at doing something cool like a terminal UI or something as well, should be as easy as calling these commands and putting them there.

delbonis:
Working on improving the `ls` command here: https://github.com/delbonis/lit/tree/issue320

PR #270, opened 2018-07-11
Made the Lit-AF ls command only print the things you tell it to.

delbonis:
Resolves the important part of #230.

### DIFF
--- a/cmd/lit-af/shell.go
+++ b/cmd/lit-af/shell.go
@@ -222,10 +222,11 @@ func (lc *litAfClient) Ls(textArgs []string) error {
 	}
 
 	if len(textArgs) == 0 {
-		fmt.Printf("pick one: conns, chans, addrs, txos, bals, ports, dualfunds")
+		fmt.Printf("pick one: conns, chans, dualfunds, txos, ports, addrs, bals")
 		return nil
 	}
 
+	// TODO Move these to their respective places?  Perhaps this gets optimized out anyways.
 	pReply := new(litrpc.ListConnectionsReply)
 	cReply := new(litrpc.ChannelListReply)
 	aReply := new(litrpc.AddressReply)
--- a/cmd/lit-af/shell.go
+++ b/cmd/lit-af/shell.go
@@ -245,75 +245,92 @@ func (lc *litAfClient) Ls(textArgs []string) error {
 		if err != nil {
 			return err
 		}
-		if len(pReply.Connections) > 0 {
+
+		if all {
 			fmt.Fprintf(color.Output, "\t%s\n", lnutil.Header("Peers:"))
+		}
+
+		if len(pReply.Connections) > 0 {
 			for _, peer := range pReply.Connections {
 				fmt.Fprintf(color.Output, "%s %s (%s)\n",
 					lnutil.White(peer.PeerNumber), peer.RemoteHost, peer.LitAdr)
 			}
+		} else {
+			fmt.Printf("none\n")
 		}
 	}
 
 	if cmd == "chans" || all {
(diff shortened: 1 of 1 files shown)


cmd/lit-af/shell.go
```diff
@@ -229,141 +235,210 @@ func (lc *litAfClient) Ls(textArgs []string) error {
 	lReply := new(litrpc.ListeningPortsReply)
 	dfReply := new(litrpc.PendingDualFundReply)
 
-	err := lc.Call("LitRPC.ListConnections", nil, pReply)
-	if err != nil {
-		return err
+	cmd := textArgs[0]
+	all := false
```
Varunram: pedantic, but could maybe change this to something like `allArgs` or something

cmd/lit-af/shell.go
```diff
@@ -229,141 +235,210 @@ func (lc *litAfClient) Ls(textArgs []string) error {
...
-	}
 
-	err = lc.Call("LitRPC.ChannelList", nil, cReply)
-	if err != nil {
-		return err
-	}
-	if len(cReply.Channels) > 0 {
-		fmt.Fprintf(color.Output, "\t%s\n", lnutil.Header("Channels:"))
+		if all {
+			fmt.Fprintf(color.Output, "\t%s\n", lnutil.Header("Peers:"))
```
Varunram: let's not print peers unless we connect like we do now?

cmd/lit-af/shell.go
```diff
@@ -229,141 +235,210 @@ func (lc *litAfClient) Ls(textArgs []string) error {
...
+					color.Output,
+					"%s (peer %d) type %d %s\n\t cap: %s bal: %s h: %d state: %d data: %x pkh: %x\n",
+					lnutil.White(c.CIdx), c.PeerIdx, c.CoinType,
+					lnutil.OutPoint(c.OutPoint),
+					lnutil.SatoshiColor(c.Capacity), lnutil.SatoshiColor(c.MyBalance),
+					c.Height, c.StateNum, c.Data, c.Pkh)
+			}
 		} else {
-			openChannels = append(openChannels, c)
+			fmt.Printf("none\n")
```
Varunram: similar to the peers comment, I think we're better off not printing channels if we don't have any like we do now. If this is too difficult, we could keep this though :)

cmd/lit-af/shell.go
```diff
@@ -221,6 +221,12 @@ func (lc *litAfClient) Ls(textArgs []string) error {
 		return nil
 	}
 
+	if len(textArgs) == 0 {
+		fmt.Printf("pick one: conns, chans, dualfunds, txos, ports, addrs, bals")
```
Varunram: should either be `fmt.Println` or you should pass `%s` and then use `fmt.Printf`

delbonis:
Ignore that second check there, that's my mistake.

delbonis:
retest this please

delbonis:
retest this please

uspv/init.go
```diff
@@ -74,45 +75,46 @@ func (s *SPVCon) GetListOfNodes() ([]string, error) {
...
 
-			s.con, err = d.Dial(conMode, conString)
+			con, err = d.Dial(conMode, conString)
 		} else {
-			s.con, err = net.Dial(conMode, conString)
+			con, err = net.Dial(conMode, conString)
 		}
 
 		if err != nil {
-			if i != len(listOfNodes)-1 {
```
Varunram: I guess this got mixed up during rebases.. https://github.com/mit-dci/lit/pull/270/commits/06417fdd2b4941cffa7686ae7b0234d3b50e7c3b doesn't contain this change

PR #271, opened 2018-07-12
Retire glit and litbamf

Varunram:
Retire litbamf and glit now that we have the webui ready. If someone wants to work on this, we could point them here and they can start off, though its highly unlikely.

Also made Jenkinsfile work with master. Demo can be viewed over at http://hubris.media.mit.edu:8080/job/jenkinsfile-test/. This PR supersedes #214 which was a standalone Jenkinsfile PR. If you'd like, I could split those changes away from this PR as well.


Jenkinsfile
```diff
@@ -0,0 +1,17 @@
...
+  stages {
+    stage('Test lit') {
+      steps {
+        sh 'rm -rf lit ../lit'
+        sh 'git clone https://github.com/Varunram/lit'
+        sh 'mv lit ../'
+        sh 'cd ..'
+        sh 'make lit lit-af'
+        sh 'make test with-python=true'
+      }
```
delbonis: Imho, this should be broken up into multiple pipeline stages.  And there should be a way to start with a "clean" workspace pointing at the commit that's being built at the beginning of the script. ([link](https://zwischenzugs.com/2017/04/23/things-i-wish-i-knew-before-using-jenkins-pipelines/)]

Varunram: well, jenkinsfile is really weird, I guess I don't want to mess around too much with it - I wanted it to work with minimal stuff, hence I didn't. But yes, can be expanded upon.

adiabat:
Sorry to see these go.  But since git is a blockchain, they are forever preserved, and one day may rise again!

PR #279, opened 2018-07-17
RPC improvements (merged Accept/Decline, detail on Fund, plain HTTP endpoint, etc.)

delbonis:
* Merged some Accept/Decline calls into generic Respond calls, could use some deeper refactoring but the RPC interface itself is cleaner now
* Added more detail to the output of the Fund call
* Added a plain HTTP endpoint for RPC at `/oneoff` so you don't have to use web~scale~sockets
* Added a "TestLog" RPC command that just writes a string to the log

Might need a little bit of testing to make sure everything works as expected still.


litrpc/listener.go
```diff
@@ -42,12 +43,40 @@ func serveWS(ws *websocket.Conn) {
 	jsonrpc.ServeConn(ws)
 }
 
+// OiOoReadWriter is for One-In-One-Out Reader/Writer and I hope to god that it works.
```
Varunram: very pedantic and nitpicking, but we could do better comments :)

Varunram:
On a side note, why would we want a http endpoint? Might as well add a https endpoint if we're adding one. I did that in an earlier PR, but it was not merged since I kicked websockets out totally :)

delbonis:
Afaik HTTPS is supported if the embedded web server supports it.  If someone really wanted it they could do it with NGINX in front of Lit and leave it at that.  But if you have access to the RPC port you probably are already in a "secured" location on the network (on the same machine or through an encrypted tunnel already), as it shouldn't be publicly-accessible.  The `ws://` is initiated over `http://`, so it's not like I'm *decreasing* the overall security.

The HTTP endpoint is really more for me since there's no JSONRPC-over-websockets libraries in Rust but there are JSONRPC-over-HTTP libraries. :laughing:

litrpc/listener.go
delbonis: It's because as far as I know Go doesn't have a thing that has a way to say "when you call `Read`, call it on this, and when you call `Write` call it on *that*".

Varunram:
Well, I'm not too sure I'd like people to mess with nginx - plus JSONRPC over HTTP should support https as well

> But if you have access to the RPC port you probably are already in a "secured" location on the network

I guess, but `ws` can be upgraded to `wss` pretty trivially, so idk.

Varunram:
As discussed today, lets have a separate document and come to consensus on the best solution

Issue #234, opened 2018-06-27
Introduce a [pay] command


gertjaap:
I'd like to suggest introducing a `pay` command in lit, with the following parameters:

`pay <cointype> <amount> <ln-address>`

What this command does now is: pay the given ln address using an off-chain payment when there's a channel to it with sufficient balance. If not, it'll exchange an on-chain address with the peer and pay it on-chain (provided we have enough in our wallet).

In the future, this could extend to "can we reach that ln address via a multi-hop route".

I think it makes for a much easier setup for payments.

PR #283, opened 2018-07-18
Initial version of the pay command

gertjaap:
This is a first implementation of the pay command described in #234 

You simply say `pay [ln address] [coin type] [amount] [data]`

The node will look for an existing channel to the given address and enough capacity. If found, it will execute a regular `push`. If this isn't found, we will ask the given node for an on-chain address and then send funds from our wallet to that on-chain address via a regular on-chain transaction.

### DIFF
--- a/cmd/lit-af/shell.go
+++ b/cmd/lit-af/shell.go
@@ -87,6 +87,12 @@ func (lc *litAfClient) Shellparse(cmdslice []string) error {
 		return parseErr(err, "send")
 	}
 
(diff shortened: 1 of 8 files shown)


gertjaap:
@metalicjames Would it make sense to update this to include your multihop command once it's merged? So that it would do a multi-hop payment if possible, and if not fall back to the on-chain payment?

gertjaap:
Will resubmit this PR once multihop is in.

PR #287, opened 2018-07-20
Have different logging tiers

Varunram:
Have 6 different logging tiers configurable with -v or --verbose. Still can't write to the logfile without the nice colors - will look into this and update


lit.go
```diff
@@ -39,8 +39,8 @@ type config struct { // define a struct for usage with go-flags
 	ReSync bool `short:"r" long:"reSync" description:"Resync from the given tip."`
 	Tower  bool `long:"tower" description:"Watchtower: Run a watching node"`
 	Hard   bool `short:"t" long:"hard" description:"Flag to set networks."`
-  
-	Verbose bool `short:"v" long:"verbose" description:"Set verbosity to true."`
+
+	LogLevel int `short:"v" long:"verbose" description:"Set verbosity level from 0 to 5 (most to least)"`
```
gertjaap: This seems somewhat backwards, i'd expect the higher the number the more logging

Varunram: Cool, will change it in the next commit.

Varunram:
retest this please

PR #289, opened 2018-07-24
Mark all regtest coins as TestCoins

Varunram:
Also make testnet3 coins not  `TestCoin`s


delbonis:
Should we have support for some sort of `LIT_FORCE_CORRECTNESS_CHECKS=1` which will make unconditional error checks happen unconditionally regardless of coins being testcoins?  Or maybe `LIT_CORRECTNESS_CHECKS={ignore,auto,force}` to apply to all coins.

Varunram:
Well, we could do that - but like these coins are testnet coins for a purpose, so I don't know :) Open to more feedback.

PR #292, opened 2018-07-25
Only show those channels whose daemons are running

Varunram:
It doesn't make sense to show these channels since this causes confusion and people would be like "why is this channel unconfirmed" when in reality its  another channel which had a bug due to insufficient funding or something.

### DIFF
--- a/cmd/lit-af/shell.go
+++ b/cmd/lit-af/shell.go
@@ -229,7 +229,11 @@ func (lc *litAfClient) Ls(textArgs []string) error {
 	lReply := new(litrpc.ListeningPortsReply)
 	dfReply := new(litrpc.PendingDualFundReply)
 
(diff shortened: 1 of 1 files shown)


gertjaap:
Could we update this PR? I would like these channels to be shown in a separate category "Disabled channels" when you do `ls chans`. 

Another thought: When trying to use these channels (since they still have an id someone could do `push <idx> ...` with the ID of this "disabled" channel, it should not be allowed. Also not if we're receiving an update. Since we cannot enforce the channels, we shouldn't allow updates to them?

Varunram:
Closing this PR since lots of changes have been made to ls since the PR was made. Lets continue this conversation in your PR :)

gertjaap:
Created #337 to track the idea for rejecting updates on "disabled" channels. Will open new PR for the LS output

Issue #291, opened 2018-07-25
Connect only to those peers associated with coin daemon(s) running


Varunram:
for e.g, if I'm running regtest and testnet3 daemons, connect only to those hosts even though I may have connected to vertcoin and testnet4 peers earlier. This ensures that stuff like wrong funding, etc doesn't happen.

A direct consequence of this is that people can't passively send you money (you can't since you don't have a daemon), but that shouldn't be a problem since people generally don't expect to connect to peers whose coin daemons they don't run. If the above behaviour is needed, we could have a receive only mode where you can connect to all peers only if explicitly desired.

Leaving this open for discussion for a couple days and will put in a PR if this idea is agreed upon.

Varunram:
PR: #293

gertjaap:
PR is merged, closing the issue

PR #293, opened 2018-07-26
Autoreconnect only to those peers attached to running coin daemon(s)

Varunram:
#291


gertjaap:
This will only reconnect peers that you have or had a channel with. The functionality used to reconnect to peers that you know, and previously connected to. For nodes you're settling a previous agreed DLC with this would mean going back to manual reconnection when settling. 

I don't see the problem of being reconnected to a peer that has/had a channel with you in a coin type you are currently not connected to a daemon for. If that is a problem, we should also disallow manual connection to such a peer.

So i don't think this is a good addition to merge in.

Varunram:
I think we can make it work with DLCs as well, haven't looked into that.

> I don't see the problem of being reconnected to a peer that has/had a channel with you in a coin type you are currently not connected to a daemon for

You can't push funds and you can't create channels, so I don't see any reason why you should except for the fact that you can receive money, which imo should be spun off into a listen only mode, where you can listen for incoming connections and channels, but can't push any funds.

gertjaap:
Discussed during the meeting: make this new logic optional under a switch

gertjaap:
Made the changes as per discussion (optional through lit.conf or command line switch option). Code will skip reconnecting if the peer does not meet the requirement of having a channel in one of the connected coin types.

PR #295, opened 2018-07-26
New ncurses UI for monitoring Lit a la htop

delbonis:
Right now it only has support for reading data from the node, adding menus to connect to nodes, open channels, make payments is coming in another PR, but there's a bit of work that needs to be done in Lit itself to make that happen first.

I also added a "LastUpdate" field to `qln.Qchan`, but for certain kinds of channel state transition it's not being updated.  It's mostly aesthetic though so it's not a lot to worry about.


adiabat:
Neato

PR #300, opened 2018-07-30
Clean up some inefficient assignments

Varunram:
Mostly trivial stuff.


qln/signtx.go
```diff
@@ -184,6 +187,9 @@ func (nd *LitNode) SignState(q *Qchan) ([64]byte, error) {
 	// generate sig.
 	bigSig, err := txscript.RawTxInWitnessSignature(
 		tx, hCache, 0, q.Value, pre, txscript.SigHashAll, priv)
+	if err != nil {
+		return sig, err
```
delbonis: Should this be `return nil, err`?

Varunram: github's  diff is misleading :) `var sig [64]byte` sits and that'd be empty when it returns here, so we're good. We can't return nil because the argument needs and a slice, so we either return `[64]byte` or we return `sig`, which in this case would be an empty slice.

gertjaap:
Fixed merge conflicts, i think this is ready to go.

Issue #244, opened 2018-06-28
Allow nodes to publish exchange rates


gertjaap:
In addition to channel states that are currently being broadcast over the network, we would need nodes to announce their exchange rate between coins in order to automatically create an atomic swap using HTLCs

Issue #243, opened 2018-06-28
Automatically generate route for forwarding payments


gertjaap:
Could be worked on after #236 and #237 are implemented.

Issue #242, opened 2018-06-28
Allow forwarding using HTLCs over a given channel path


gertjaap:
Could be worked on after #236 and #237 are implemented.

gertjaap:
Implemented in #305

PR #305, opened 2018-08-02
Multi-hop payments

metalicjames:
This PR adds multi-hop payment support to Lit. It is possible to route from any origin coin type to a different coin type, thus enabling cross-chain swaps. Routing fees can be implemented by setting an exchange rate between the same coin type other than 1.

Adds the `paymultihop` command to `lit-af` e.g: 
`paymultihop ln1up6k8vxt7vjqcnyzq08e5w6gv8ckr466fcswl3 28 1 250000` 
to pay `ln1up6k8vxt7vjqcnyzq08e5w6gv8ckr466fcswl3` 250000 satoshis of coin type 1 to be delivered as coin type 28.

At present preimages are generated randomly.


treyzania:
>At present preimages are generated randomly and are not saved between restarts. That should probably be rectified before this PR is merged.

As in they're not serialized to disk?

metalicjames:
That is correct.

qln/break.go
```diff
@@ -40,6 +40,21 @@ func (nd *LitNode) BreakChannel(q *Qchan) error {
...
 	// set delta to 0... needed for break
 	q.State.Delta = 0
+
+	for i, h := range q.State.HTLCs {
+		if !h.Cleared && h.Clearing {
+			q.State.HTLCs[i].Clearing = false
+		}
+	}
+
+	if q.State.InProgHTLC != nil {
```
delbonis: You don't need to do a branch here, if it's already nil then you won't do anything.

delbonis:
@metalicjames There's a few merge conflicts here, can you take a look at them?

gertjaap:
Implements #242 #243 #244

PR #307, opened 2018-08-03
Connect to known nodes via Noise_XK

Varunram:
Since we know the pubkey of nodes we connected to previously, use Noise_XK to connect to them and connect to new nodes using Noise_XX


delbonis:
Should we have some way for forcibly connect to a known node using XX?

Varunram:
Hm, to break the trust web that the PK provides, nice idea, maybe I'll put that into its own PR so that it doesn't delay this PR getting in :) Might want to have that option in `lit` itself to ensure that we don't trust anyone by their PKs permanently.

adiabat:
This is cool -- not sure why you'd want to force XX though if you already have the key?  It seems like the only advantage of XX is that you learn the remote key at the end.

Varunram:
Thinking back, I don't know either. If he has a new pubkey, he'd have a new listening address as well. But might be that you want to forcibly connect because you want to reverify his PK or something..

adiabat:
XK mode does (re?)verify their pubkey.  It just doesn't tell you it if you don't know it already.

lndc/conn.go
```diff
@@ -70,22 +86,30 @@ func Dial(localPriv *btcec.PrivateKey, ipAddr string, remotePKH string,
...
+		if _, err := b.noise.RecvActTwo(actTwo); err != nil {
+			b.conn.Close()
+			return nil, err
+		}
 	}
-
-	logging.Info("Received pubkey", s)
-	if lnutil.LitAdrFromPubkey(s) != remotePKH {
+	logging.Infoln("Received pubkey", remotePK)
+	if lnutil.LitAdrFromPubkey(remotePK) != remotePKH && !Noise_XK {
```
gertjaap: Seems to make sense to switch these clauses around, so that if Noise_XK is true it will not do the comparison.

lndc/conn.go
```diff
@@ -70,22 +86,30 @@ func Dial(localPriv *btcec.PrivateKey, ipAddr string, remotePKH string,
...
+		// for noise_XK dont check PKH and PK because we'd have already checked this
+		// the last time we connected to this guy
 		return nil, fmt.Errorf("Remote PKH doesn't match. Quitting!")
 	}
-	logging.Infof("Received PKH %s matches", lnutil.LitAdrFromPubkey(s))
+	logging.Infof("Received PKH %s matches", lnutil.LitAdrFromPubkey(remotePK))
```
gertjaap: This logging statement should be different when Noise_XK is true - you didn't check that it matches. Something like "Assuming PKH still matches" or just don't log it at all.

Varunram:
Closing this in favour of #435

Issue #275, opened 2018-07-15
Add support for batching connections


Varunram:
Right now, we collect a list of peers by pinging the DNS seeds and then try to connect to them one by one. What we can do to speed this up is to try connecting to say 5 peers at once and then connect to the one which responds first (and drop the other 4 peers). Something more advanced could be to keep the 4 peers in an array (arranged by increasing TTLs) and if the peer we connected to disconnects, we can immediately dial up the next one on the list. A natural extension of this would be to support connecting to multiple peers at once.

Varunram:
Waiting for noise_XX to be final before starting on this to avoid merge conflicts and rebases.

Varunram:
PR: #324 (#313 was closed due to a high probability that lit would get banned)

PR #313, opened 2018-08-08
Batch connections on startup

Varunram:
On startup, we only  connect to a single node and if it fails, we proceed to dial the next node in `listOfNodes`. Instead, we can batch connections and open parallel connections with all of them and simply drop a peer if it doesn't respond / does something weird.

This is also configurable via a command line option `maxConnections` or `-m`. Issue: #275

### DIFF
--- a/uspv/init.go
+++ b/uspv/init.go
@@ -135,7 +135,6 @@ func (s *SPVCon) DialNode(listOfNodes []string) error {
 		}
 	}()
 	wg.Wait()
-	s.conns[0] = s.conns[0] // remove this
 	return nil
 }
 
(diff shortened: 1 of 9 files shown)


uspv/init.go
```diff
@@ -108,15 +125,24 @@ func (s *SPVCon) DialNode(listOfNodes []string) error {
 				return fmt.Errorf(" Tried to connect to all available node Addresses. Failed")
 			}
 		}
-		break
 	}
+	go func() {
```
adiabat: why is this in a goroutine...?

Varunram:
retest this please

Varunram:
The right way to do this would be to have concurrent connections since dropping connections often would end up in lit nodes getting banned by bitcoin nodes.

Varunram:
Closed in favour of #324 which does concurrent connections.

PR #316, opened 2018-08-08
Rewrote integration tests with new test framework

delbonis:
The tests now work on more specific behavior and that helps you find where bugs are more easily and find edge cases better.

I haven't written any negative tests ("this *shouldn't* work"), but that should be pretty trivial now.


delbonis:
So it makes sense that the Jenkins tests are failing because it's set up to run the old tests.

delbonis:
Ok should be good to go now.

delbonis:
The `close` test(s) don't check to see if the other end knows the channel has closed.  Presumably I forgot to do this in the `break` tests too.  I'll update that soon.

PR #317, opened 2018-08-10
Added missing AcceptOrDecline argument

navybluesilver:
Contracts are always being declined when using the accept command. This is because the AcceptOrDecline argument is missing.

### DIFF
--- a/cmd/lit-af/dlccmds.go
+++ b/cmd/lit-af/dlccmds.go
@@ -878,6 +878,7 @@ func (lc *litAfClient) dlcContractRespond(textArgs []string, aor bool) error {
 	}
 
 	args.CIdx = cIdx
+	args.AcceptOrDecline = aor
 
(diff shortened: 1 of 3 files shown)


Varunram:
retest this please

Varunram:
I think this is due to the many rebases  we’ve had, thanks!

delbonis:
Yeah maybe we should consider writing tests for `lit-af` itself?

And the reason the argument is named `aor` is because it referred to "Accept or Reject" in other places, perhaps we should rename things to be more internally-consistent.

Varunram:
This is actually because we didn't follow set standards for merging of PRs. Please refer to my comment at #319.

infografx:
Hello guys. Could you tell me please, why there is no "DeleteContract" function in the master branch.

Should some work be done in this direction?

PR #324, opened 2018-08-18
Concurrent Connections on Startup

Varunram:
Connect simultaneously to multiple peers and send / receive messages from them. An extension of Batch Connections which had a problem that peers would try to ban us. Default maxConnections value is 5 and the maximum maxConnections value is 120. Works fine as tested, although there may be some stuff I missed out. Nodes also have an additional identifier peerIdx based which the message streams are handled.


gertjaap:
retest this please

PR #326, opened 2018-08-20
[WIP] Event bus system, replacing ad-hoc channel-based notification systems

delbonis:
I've noticed in some places we're basing triggering certain actions around sending a message to a channel.  This works, but it's not very extensible and it leads to some bad habits.

This kind of event bus is just a simple design that uses strings as event names and an interface for the overall "`Event`" type.  Events be handled synchronously or asynchronously on a per-event basis, and there's logic to "cancel" and even if-needed.  There isn't a "handler priority" system but that shouldn't be too difficult to develop.

The next steps for this is just porting over old code to use this event bus system and adding events for different kinds of actions.

### DIFF
--- a/eventbus/bus.go
+++ b/eventbus/bus.go
@@ -1,19 +1,23 @@
 package eventbus
 
-import "sync"
-import "fmt"
+import (
+	"fmt"
+	"sync"
+)
 
 // An EventBus takes events and forwards them to event handlers matched by name.
 type EventBus struct {
-	handlers map[string][]*eventhandler
-	mutex    *sync.Mutex
+	handlers     map[string][]*eventhandler
+	eventMutexes map[string]*sync.Mutex
+	mutex        *sync.Mutex
 }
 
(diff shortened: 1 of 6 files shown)


Varunram:
What's the proposed advantage of using the events system as opposed to channels? And how would we return values from channels and do all the traditional stuff that we do with channels right now?

delbonis:
It's difficult to have multiple things respond to a value sent on a particular channel.  This isn't a 1:1 replacement but for situations where there isn't a tight logical coupling between sender and receiver(s) it's much clearer to understand.  Also it makes extending things much clearer as there's nothing needed in the sender to add more responding logic.

delbonis:
If we actually have a *stream* of data that we're moving then it makes sense to use a channel.  But if we're using the channel to signal other threads/goroutines about something then it's not a great idea to use a channel.

delbonis:
Branch merged into [`refactor/dbandp2p`](https://github.com/delbonis/lit/tree/refactor/dbandp2p).

PR #329, opened 2018-08-26
Lit Invoice Protocol

Varunram:
Not complete yet, but good enough for basic testing and functionality. Right now, it still uses underscore as a separator, need to change that. Need to store and generate invoices as well - right now its hardcoded. Also need to do routing / multi hop stuff once that goes into master. What it does right now is push funds through an existing channel if it exists and open a new channel if a channel doesn't already exist and push funds through that. Also, need a good name (LIP?)


gertjaap:
retest this please

Issue #350, opened 2018-09-12
Starting lit with invalid daemon name doesn't work


Varunram:
Before the recent mer gets, we could still start `lit` with invalid params. Now that doesn't seem possible and we crash out. Will fix along with the other bugs reported.

delbonis:
Isn't that a good thing?

Varunram:
Good and bad actually. If this is good then lit must quit when it loses connection to a remote peer. If this is bad, we must fix this :) My view is that its good because we want behaviour in both cases to be identical (like it was earlier), hence the issue.

Varunram:
Fixed in #372

Issue #349, opened 2018-09-11
Lit doesn't start


Varunram:
With recent merges, lit doesn't seem to start anymore

```
2018/09/12 00:21:09.497789 [WARN] WARNING!! Key file not encrypted!!
2018/09/12 00:21:09.498155 [WARN] Anyone who can read the key file can take everything!
2018/09/12 00:21:09.498193 [WARN] You should start over and use a good passphrase!
2018/09/12 00:21:09.503528 failure loading exchange rates: unexpected end of JSON input
2018/09/12 00:21:09.576294 [ERROR] Error loading StatCom: EOF
2018/09/12 00:21:09.576330 EOF
```

It shouldn't ideally fail at `rates.json` and create one with standard values, if possible. Also, having an empty `rates.json` file still gives the same output (`Error loading StatCom: EOF`)

Varunram:
Closed via 
https://github.com/mit-dci/lit/pull/353/commits/2201e76b3511202946305f36b3f8c9ad7589cb57

Issue #348, opened 2018-09-11
Help command doesn't work


Varunram:
`./lit -h` results in a grouting crash
```
goroutine 1 [running]:
main.litSetup(0xc4201ba000, 0x0)
	/Users/varunramg/go/src/github.com/mit-dci/lit/litinit.go:100 +0x9ac
main.main()
	/Users/varunramg/go/src/github.com/mit-dci/lit/lit.go:174 +0x172
```
It also prints the help output twice. Will put in a PR to fix this.

Varunram:
Fixed in recent merges.

PR #330, opened 2018-08-29
Refactoring P2P setup code and set up new DB layer framework

delbonis:
This PR replaces the peer connection code with a new event-based system.  It's not complete and there's some things that will be changed in the future.  Some changes will be made in a future PR as they're waiting on another thing to happen.  A rebase will be needed after the new integration test framework is merged.

TODO:
- [x] Connecting to peers, listening for connections
- [x] Fully persistent peer ids
- [x] Disconnecting
- [ ] ~~DB stuff~~
    - [x] Storing/loading `PeerInfo`
    - [ ] ~~Wallet database (outpoints to watch, spendable utxos, etc.)~~
    - [ ] ~~Channel info~~
    - [ ] ~~DLC stuff~~
- [x] Connecting through a proxy
- [x] NAT holepunching, etc.

*(Crossed out items are likely postponed until other refactors have been done.)*

PR #351, opened 2018-09-12
[WIP] Regression fixes

Varunram:
A single PR to fix all regressions reported in issues #348, #349, #350 (and more to come).

EDIT: Added WIP to denote running status (as bugs are reported, close them in this PR)

### DIFF
--- a/litinit.go
+++ b/litinit.go
@@ -2,13 +2,10 @@ package main
 
 import (
 	"bufio"
-	"fmt"
 	"io"
 	"log"
 	"os"
 	"path/filepath"
-
-	"github.com/mit-dci/lit/logging"
 
 	"github.com/jessevdk/go-flags"
 	"github.com/mit-dci/lit/lnutil"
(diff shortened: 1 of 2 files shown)


delbonis:
Wait until after #330 so that I don't have to spend another hour rebasing.

Varunram:
I think it doesn't make sense to push more changes when we have regression on master. Lets wait for regression fixes (should be only a few lines and no major merge conflicts) and then test #330 effectively.

litinit.go
```diff
@@ -60,28 +57,28 @@ func litSetup(conf *config) *[32]byte {
 	// create home directory
 	_, err = os.Stat(preconf.LitHomeDir)
 	if err != nil {
-		fmt.Println("Error while creating a directory")
```
gertjaap: The log file is not yet configured there. So using log and fmt both write to stdout (it will only write to stdout). Right? But it adds a package import.

Varunram: a package import vs a panic every time we print out the help command

Varunram:
Improved upon in other PRs. Closing this one.

PR #353, opened 2018-09-13
Logging fixes

gertjaap:
- Implemented the new logging system into new code that was merged
- Implemented the use of `logging` into lit-af
- Merged Panic and Fatal log levels
- Calls to `Fatal...` will always execute, since it's level 0 now.
- Fixed some places where panic() was used in stead of logging.Fatal 
- Removed the use of `log` everywhere but in the logging package.


gertjaap:
Addresses issues #348 #349 and #350

PR #362, opened 2018-09-17
Don't work on a copy of the config struct

Varunram:
Change the config working part where we were working on a copy of the config instead of the actual one.


Varunram:
retest this please

PR #368, opened 2018-09-18
Allow connecting to coindaemon peers by hostname, not just localhost

gertjaap:
This allows you to connect to coindaemons not just by IP or `localhost` but also hostname. So i could use `reg=someserver:18444`


delbonis:
This is very useful for the vending machine.

PR #370, opened 2018-09-18
Just remove obsolete logging statements. Elkrem not needed for break

gertjaap:
This fixes the failing two tests breaking right after fund. The elkrem is empty, but also not needed to break. Elkrem is needed for justice - at state 2 or higher.

Return value of ElkRcv is not used, but it used to crash this section.

Re-enabled the two tests


adiabat:
yay :)

PR #384, opened 2018-09-21
Fixed peerIdx confusion after reconnecting/restarting nodes, with tests

delbonis:
I don't think that I need to restart the node that's reconnecting to the other nodes, but that was the simplest idea I had to make sure that the nodes would actually disconnect.

I also had to refactor a bit of the testlib code to get it to support dynamically starting/shutting down nodes on demand.

### DIFF
--- a/lnp2p/peermgr.go
+++ b/lnp2p/peermgr.go
@@ -118,13 +118,11 @@ func computeIdentKeyFromRoot(rootkey *hdkeychain.ExtendedKey) (privkey, error) {
 }
 
 // GetPeerIdx is a convenience function for working with older code.
-func (pm *PeerManager) GetPeerIdx(peer *Peer) int32 {
-	for i, p := range pm.peers {
-		if pm.peerMap[p] == peer {
-			return int32(i)
-		}
+func (pm *PeerManager) GetPeerIdx(peer *Peer) uint32 {
+	if peer.idx == nil {
+		return 0
 	}
-	return 0
+	return *peer.idx
 }
 
 // GetPeer returns the peer with the given lnaddr.
--- a/lnp2p/peermgr.go
+++ b/lnp2p/peermgr.go
@@ -23,6 +23,10 @@ import (
 type privkey *koblitz.PrivateKey
 type pubkey *koblitz.PublicKey
 
+// MaxNodeCount is the size of the peerIdx->LnAddr array.
+// TEMP This shouldn't be necessary.
+const MaxNodeCount = 1024
+
 // PeerManager .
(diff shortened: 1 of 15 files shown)


delbonis:
I'm not sure if this PR needs more work or if it's good enough as-is.  I think we should wait until after the PR that fixes the bug this is testing is merged before merging this so we don't "break" master with these new tests.

Varunram:
One test fails and 6 are ignored:
```
23:01:10 Failures:
23:01:10  - reconnect:reordered
```

test/runtests.py
```diff
@@ -64,27 +64,29 @@ def load_tests_from_file(path):
             modname = 'testmod_' + tname
             mod = __load_mod_from_file(fname, modname)
             mods[tname] = mod
+        pretty = tname
+        tfn = getattr(mod, t['func_name'])
```
Varunram: nitpicking, but what's `tfn` here?

test/itest_reconnect.py
```diff
@@ -0,0 +1,48 @@
...
+
+def reordered(env):
+    print('Starting nodes')
+    lit1 = env.new_lit_node()
+    lit2 = env.new_lit_node()
+    lit3 = env.new_lit_node()
+    print('Connecting nodes... (2 then 3)')
+    lit1.connect_to_peer(lit2)
+    lit1.connect_to_peer(lit3)
+    l2p1 = lit1.get_peer_id(lit2)
```
Varunram: Again, the test looks good, but we could do with better names since it would be easier for people to understand this in the future.

test/runtests.py
delbonis: Test Function Name

test/itest_reconnect.py
delbonis: yeah I guess I could comment that

delbonis:
This now actually fixes the bug the tests were testing.

lnp2p/peermgr.go
```diff
@@ -118,13 +118,11 @@ func computeIdentKeyFromRoot(rootkey *hdkeychain.ExtendedKey) (privkey, error) {
...
-	for i, p := range pm.peers {
-		if pm.peerMap[p] == peer {
-			return int32(i)
-		}
+func (pm *PeerManager) GetPeerIdx(peer *Peer) uint32 {
+	if peer.idx == nil {
+		return 0
 	}
-	return 0
+	return *peer.idx
```
Varunram: Wonder how this got in, nice catch!

delbonis:
retest this please

PR #387, opened 2018-09-23
Ports should be of type int

Varunram:
Doesn't make sense passing around a port with a colon, just pass the port and colonize where necessary. Also maybe need a constant for the default port..


Varunram:
retest this please

Varunram:
retest this please

PR #400, opened 2018-09-27
Allow closing LNDC RPC client

gertjaap:
When you use an LndcRpcClient, you can't close the connection. While building some utilities that interface with lit this becomes kind of problematic.


adiabat:
looks straightforward enough

PR #403, opened 2018-09-29
Restructure the way stuff is handled in lnp2p

Varunram:
Contains name changes, call by value / reference changes, removes unnecessary handler functions and more.


Varunram:
test fails due to jenkins being down, unrelated to the PR itself.

db/lnbolt/peerdb.go
```diff
@@ -64,51 +64,46 @@ func (pdb *peerboltdb) GetPeerAddrs() ([]lncore.LnAddr, error) {
...
+func (pdb *peerboltdb) GetPeerInfo(addr string) (lncore.PeerInfo, error) {
 
 	var raw []byte
 	var err error
+	var peerInfo lncore.PeerInfo
 
 	if pdb.db == nil {
 		logging.Warnf("PDB.db is nil!")
-		return nil, fmt.Errorf("PDB.db is nil")
+		return peerInfo, fmt.Errorf("PDB.db is nil")
```
delbonis: This doesn't really make sense to me because you're still returning *a* peerinfo, but it's invalid.  Sure if `err` is non-nil then you should ignore it anyways but if you mess up and forget then you're going to run into issues if the peerinfo isn't a null pointer.

lncore/channels.go
```diff
@@ -16,7 +15,7 @@ type LitChannelStorage interface {
 // ChannelHandle is "something" to concisely uniquely identify a channel.  Usually just a txid.
 type ChannelHandle [64]byte
 
-// ChannelState .
+// ChannelState
```
delbonis: I left the `.` in because `go vet` was yelling at me if I didn't have it because it wanted "properly formatted documentation" but since it isn't fully-defined yet it doesn't make sense to have full docs for that.

lncore/peers.go
```diff
@@ -1,21 +1,17 @@
...
+	DeletePeer(addr string) error
 }
 
-// PeerInfo .
+// PeerInfo stores peer related data
 type PeerInfo struct {
-	LnAddr   *LnAddr `json:"lnaddr"`
-	Nickname *string `json:"name"`
-	NetAddr  *string `json:"netaddr"` // ip address, port, I guess
+	Addr     string `json:"string"`
```
delbonis: Why get rid of `LnAddr`?

lnp2p/listen.go
```diff
@@ -59,29 +60,26 @@ func acceptConnections(listener *lndc.Listener, port int, pm *PeerManager) {
 		}
 
 		// Add the peer data to the DB if we don't have it.
-		if pi == nil {
+		if len(pi.Addr) == 0 { // weird hack, but should work
```
delbonis: The reason I don't like having this again is because then the logic for "is the lnaddr defined?" be to check the length as a string, but if we end up making LN addrs more complicated than just a plain primitive then we have a lot of work to do to change that code later on.

lnp2p/msg.go
```diff
@@ -7,7 +7,7 @@ import (
 
 // FIXME This is a stub function that just calls out to the lnutil lib for later.
 func processMessage(b []byte, peer *Peer) (lnutil.LitMsg, error) {
-	m, err := lnutil.LitMsgFromBytes(b, peer.GetIdx())
+	m, err := lnutil.LitMsgFromBytes(b, peer.Idx)
```
delbonis: This also makes things annoying to change later because if we change how Idxes are stored then we have tons of code reaching directly into the Peer datastructure to get that information instead of calling a function that we can change the implementation of more easily.

lnp2p/msg.go
```diff
@@ -57,7 +57,7 @@ func sendMessages(queue chan outgoingmsg) {
 		// Make sure the connection isn't closed.  This can happen if the message was queued but then we disconnected from the peer before it was sent.
 		conn := recv.peer.conn
 		if conn == nil {
-			logging.Warnf("peermgr: Tried to send message to disconnected peer %s\n", recv.peer.GetPrettyName())
+			logging.Warnf("peermgr: Tried to send message to disconnected peer %d, %s\n", recv.peer.Idx, recv.peer.Nickname)
```
delbonis: If `Nickname` isn't defined then it doesn't show anything, `GetPrettyName` returns a truncated version of the addr if there's no nickname defined.

lnp2p/peer.go
```diff
@@ -8,66 +8,20 @@ import (
...
-	}
-}
-
-// GetLnAddr returns the lightning network address for this peer.
-func (p *Peer) GetLnAddr() lncore.LnAddr {
-	return p.lnaddr
-}
-
-// GetRemoteAddr does something.
+// GetRemoteAddr returns the remote node's public key.
```
delbonis: oops I forgot to put a better description in there

lnutil/litadr.go
```diff
@@ -71,6 +72,13 @@ func LitAdrFromPubkey(in [33]byte) string {
 	return bech32.Encode("ln", doubleSha[:20])
 }
 
+// Given a koblitz pubkey, returns the lit addr
+func ConvertPubkeyToLitAddr(in *koblitz.PublicKey) string {
```
delbonis: We're supposed to be taking things *out* of lnutil, not putting them back in...

lnp2p/peermgr.go
```diff
@@ -141,11 +137,12 @@ func (pm *PeerManager) GetPeerByIdx(id int32) *Peer {
 	return pm.peerMap[pm.peers[id]]
 }
 
-// TryConnectAddress attempts to connect to the specified LN address.
+// TryConnectAddress is a handler function for tryConnectPeer and
```
delbonis: I'm not sure I understand what you mean by "is a handler function" here.  This tries to hide the details of calling `tryConnectPeer`.

lnp2p/peer.go
```diff
@@ -8,66 +8,20 @@ import (
 
 // A Peer is a remote client that's somehow connected to us.
 type Peer struct {
-	lnaddr   lncore.LnAddr
-	nickname *string
-	conn     *lndc.Conn
-	idpubkey pubkey
+	Addr     string
+	Idx      uint32 // deprecated, default value for all uint32s are 0
+	Nickname string
```
delbonis: I don't like having these things exported since it makes it much more difficult to change the implementation later.

qln/msghandler.go
```diff
@@ -14,49 +14,47 @@ import (
 
 func (nd *LitNode) registerHandlers() {
 
-	mp := nd.PeerMan.GetMessageProcessor()
```
delbonis: Is there a reason we can't just have a reference on the stack to the mproc?  `DefineMessage` calls it on a reference already anyways.

delbonis:
Don't mean to frustrate you with all the comments, I just made lnp2p very opinionated because I wanted to make it easier to refactor around.

db/lnbolt/peerdb.go
Varunram: While calling this function, as always is general practice in go, you catch the error first before using the returned `peerInfo` struct. This doesn't change that behaviour and we do this in other places as well.

lncore/channels.go
Varunram: Oops, yeah, we need a formatting guideline or something. But I can add this back in if go vet complains.

lncore/peers.go
Varunram: We must define the stuff in the function definition as `addr string` instead of defining a type `LnAddr` and then calling `addr LnAddr`. Also, I don't see the type of the address (string) changing for now, so I don't see why we should have a redundant type when we just can use `string`.

qln/msghandler.go
Varunram: There is no pass by reference in go :)

lnp2p/peer.go
Varunram: Yeah, I thought about this for a while as well, but we have function maps like `instance.GetVal()` for things which should be as easy as `instance.Val`. If we're using the accessor function everywhere like we do right now, we might as well export the value itself.

lnp2p/peermgr.go
Varunram: Hm, I don't think this is a "strict" change, so I can remove this

lnutil/litadr.go
Varunram: `lnutil` is supposed to be a collection of such utility functions unless I'm missing something here.

lnp2p/msg.go
Varunram: Ah, this is another debate in itself. The reason why I changed it is because of users themselves. Say users give the same pretty name to both, you don't want one guy masquerading as another, etc. But again, not a "strict" change by itself, so can add it back in.

lnp2p/msg.go
Varunram: `GetIdx()` just returned the Idx earlier. `peer.Idx` gets it directly. Again, using an accessor function makes sense _if_ you're doing something complex, but this function used to return `peer.Idx` anyway, hence the removal of that function. In the future, if we do something with peer.Idx that warrants it to be accessed in a different way than it is right now, we can use an accessor function then. Just that it doesn't have utility now :)

lnp2p/listen.go
Varunram: I think as mentioned, we won't (and shouldn't) make LnAddrs more complex than a string because a ln address is just a string :) We do have different data structures in place that can handle abstractions off LnAddr (plain `string` in this PR), so I don't think that'd be a problem.

Regarding the comment itself, I don't like the hack myself, but it works for what it intends to do, hence my choice. Open to alternatives as well.

lnp2p/msg.go
delbonis: The point is that if we have to change it to be something nontrivial then the only thing we have to change *is* the function.

qln/msghandler.go
delbonis: We're directly reaching deeply into structs again here which makes things complicated to refactor.  In this case it doesn't matter since I don't see how we keep track of the message processor changing in the future but imho it looks messier to do the whole `nd.PeerMan.MProc` dance.

But also pass by reference (done differently here):
```go
func (thing *Thing) SomeFunc(obj *OtherThing) {
}
...
var t Thing
var o OtherThing
t.SomeFunc(&o)
```

lnp2p/listen.go
delbonis: This is hard because Go's type system is just so limited it's hard to encode semantics into the type system in a nice way.

lncore/peers.go
delbonis: Really it makes sense to use `string` externally but then internally use `LnAddr`.  With the idea that we're enforcing if the type is a `LnAddr` then it *must* be a valid address and if it's just a `string` then we haven't verified that yet.  Also it being a different type stops people from directly modifying it without trying.

lnp2p/msg.go
delbonis: Well if nicknames are only local labels it shouldn't matter?  It's your own fault if you give the same nickname to two nodes.

lnp2p/msg.go
Varunram: As I said, I don't think `peer.Idx` (or any field that is exported) will change because we have handlers higher up that do other stuff. Some fields are still unexplored because we expect them to change :)

lnp2p/msg.go
Varunram: Agreed, the idea is that there must atleast be a log fallback (hence the change) so that they understand it later (and makes it easier for them to debug). I personally don't like nicknames for nodes, so this view could be a result of that as well, so I'm open to revert if others feel appropriate.

lncore/peers.go
Varunram: If the type is a string, it could be an address (valid or not). There should be appropriate functions to check if its valid or not and that mustn't move on to the types themselves (since that encourages bad practices wrt code quality). Something like `string.IsValidAdr()` would work.

Reg. people using it, I do agree we need a schema of sorts to prevent people from changing stuff. If people can change `addr string` to `addr blah`, they could change `addr LnAddr` to `addr blah` equally easily. I'm saying that we need to fix the root problem, which is not having handlers for all types that we use and that a new type set doesn't change that. I'll put the handlers in a follow-up PR once this is merged :)

qln/msghandler.go
Varunram: Yeah, we could do the pointer hack, but its not something that the go pundits recommend, idk

lncore/peers.go
delbonis: >that encourages bad practices wrt code quality

I beg to differ.  I think that actually encourages good code quality because as long as you don't intentionally break invariants you can rely on the types to encode the semantics of the data (very common in pure-FP languages).  It also "hides" behavior (like passing a `LnAddr` to things in the `strings` package) that you don't need anymore and stops you from doing "bad" things.

Granted, there's nothing stopping you from doing `x := lncore.LnAddr(strings.SomethingBad(string(x)))` but there's nothing we can do about that.

lnp2p/msg.go
delbonis: Perhaps we should have a `ResolvePeer(string) -> *Peer` function on `PeerManager` that takes looks up a literal address or falls back to searching nicknames if that fails.

lnutil/litadr.go
delbonis: Yes but it's becoming a conglomeration of a bunch of things and it's difficult to do anything without pulling in lnutil.  To be fair this is the kind of thing that doesn't matter that much to be in there, I was more annoyed with the numerous `*lib.go` files that should be in better places.

qln/msghandler.go
delbonis: that's not a hack...

adiabat:
There are several large binary files in this PR, can you remove them (and maybe rebuild the git commit log to never have them so they're not included in the history)

Varunram:
should be good now

PR #405, opened 2018-10-02
Minor changes to gendepgraph.sh and Makefile

delbonis:
Makes them respect some of the changes we do with `build/env.sh` and making things generally cleaner.


Makefile
```diff
@@ -33,7 +33,7 @@ clean:
 	go clean .
 	go clean ./cmd/lit-af
 	rm -rf build/_workspace/
-	rm -f cmd/lit-af/lit-af
+	rm -f lit cmd/lit-af/lit-af lit-af
```
Varunram: I think the last `lit-af` can be removed since that doesn't get built in `/`?

scripts/gendepgraph.sh
```diff
@@ -6,4 +6,5 @@ if [ -n "$1" ]; then
 	out=$1
```
Varunram: Do we really want the dependency graph to be present as an image file at `/`? This would mean that every time we refactor something, we'd need to update this file as well. It'd be nice if we could leave this to the user and add a note in the readme about the function of the scripts present in `scripts/`

Makefile
delbonis: It does when you build with `make`.

scripts/gendepgraph.sh
delbonis: The default is to put it in `docs`.  I had it as an image in the readme since I've seen that in a lot of Go projects.

PR #406, opened 2018-10-03
Added more graceful handling to some errors on startup

delbonis:
We don't complain about perfectly normal things now, like not finding the config file.  I also added an envvar called `LIT_KEYFILE_WARN` that if `0` will suppress the warning about having an unencrypted `privkey.hex`.

### DIFF
--- a/litinit.go
+++ b/litinit.go
@@ -71,11 +71,8 @@ func litSetup(conf *litConfig) *[32]byte {
 		}
 	}
 
-	if _, err := os.Stat(filepath.Join(filepath.Join(conf.LitHomeDir), "lit.conf")); os.IsNotExist(err) {
+	if _, err := os.Stat(filepath.Join(conf.LitHomeDir, "lit.conf")); os.IsNotExist(err) {
 		// if there is no config file found over at the directory, create one
-		if err != nil {
-			fmt.Println(err)
-		}
 		logging.Infof("Creating a new config file")
 		err := createDefaultConfigFile(filepath.Join(conf.LitHomeDir)) // Source of error
 		if err != nil {
@@ -83,7 +80,7 @@ func litSetup(conf *litConfig) *[32]byte {
 		}
 	}
 
-	conf.ConfigFile = filepath.Join(filepath.Join(conf.LitHomeDir), "lit.conf")
+	conf.ConfigFile = filepath.Join(conf.LitHomeDir, "lit.conf")
 	// lets parse the config file provided, if any
 	err = flags.NewIniParser(parser).ParseFile(conf.ConfigFile)
 	if err != nil {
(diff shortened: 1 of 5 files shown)


litinit.go
```diff
@@ -71,19 +71,16 @@ func litSetup(conf *litConfig) *[32]byte {
...
-			fmt.Println(err)
-		}
 		logging.Infof("Creating a new config file")
 		err := createDefaultConfigFile(filepath.Join(conf.LitHomeDir)) // Source of error
 		if err != nil {
 			logging.Fatal(err)
 		}
 	}
 
-	conf.ConfigFile = filepath.Join(filepath.Join(conf.LitHomeDir), "lit.conf")
```
Varunram: No idea how I didn't catch this.

lnutil/keyfile.go
```diff
@@ -60,9 +60,13 @@ func LoadKeyFromFileArg(filename string, pass []byte) (*[32]byte, error) {
 	}
 
 	if len(enckey) == 32 { // UNencrypted key, length 32
-		logging.Warnf("WARNING!! Key file not encrypted!!\n")
-		logging.Warnf("Anyone who can read the key file can take everything!\n")
-		logging.Warnf("You should start over and use a good passphrase!\n")
+		v, p := os.LookupEnv("LIT_KEYFILE_WARN")
```
Varunram: I think env vars are bad and no one is going to set them anyway. I don't see why we need to suppress this error. Lets print it out and if the user wants to act on it, he can. The logging level is on `Warning` anyway, so they can suppress it if they want to by setting an appropriate log level.

qln/routing.go
```diff
@@ -576,7 +576,11 @@ func (nd *LitNode) PopulateRates() error {
 
 	jsonFile, err := os.Open(ratesPath)
 	if err != nil {
-		return err
+		if !os.IsNotExist(err) {
+			return err
+		}
+		logging.Infof("Rates file not found.")
+		return nil
```
Varunram: We must return errors to the handler above, so something like return `return fmt.Errorf("")` should be fine. The info log helps as well, although, it could be marked `Debug` (since we have tons of `Info` logs)

delbonis:
retest this please

lnutil/keyfile.go
delbonis: Right now as far as I know there's no way for the user to generate an unencrypted keyfile, so if it's encrypted then they definitely did it intentionally and know what they're doing.  I had this so that when I'm running the test framework with `LIT_OUTPUT_SHOW=1` we decrease the noise that's printed.  (That envvar that's picked up by the testing framework.)

Really a lot of these should be config options but I don't want to make them CLI args since that's even more intrusive.  So we should push setting that up to if/when we have a config that's more than just another place for CLI args.

qln/routing.go
delbonis: Not finding the rates file isn't an error.  Although I guess I should put that `!os.IsNotExist(err)` in the caller.

delbonis:
Rebased.

PR #407, opened 2018-10-05
Added support for p2p function calls

delbonis:
This PR introduces a new concept of p2p function calls.  Currently uses msgids `0xC0`, `0xC1`, and `0xC2`.

Hopefully this will help us transition all of the msgids that are used to fake function-like calls to other peers to instead just be functions.  See below:

```go
MSGID_DLC_OFFER               = 0x90 // Offer a contract
MSGID_DLC_ACCEPTOFFER         = 0x91 // Accept the contract
MSGID_DLC_DECLINEOFFER        = 0x92 // Decline the contract
```

With this new system it's just a single 16-bit "function ID", and there's fewer barriers to having multiple operations in-progress since it operates with a system of callbacks.

```go
msg := ...
peer.InvokeAsyncCall(msg, func(resp PeerCallMessage, err error) (bool, error) {
    // handle message/error
    return true, nil
})
```

There's also support for (configurable) timeouts and a blocking version of the call interface, see the bottom of `lnp2p/peer.go`.

Currently, this PR strictly adds behavior, it doesn't remove or (significantly) change anything.

~~Still need to write some tests for this PR, as well.~~ Tests have been written.

### DIFF
--- a/lnp2p/callmsgs.go
+++ b/lnp2p/callmsgs.go
@@ -1,0 +1,100 @@
+package lnp2p
+
+import (
+	"bytes"
+	"encoding/binary"
+)
+
+const (
+
+	// MsgidCall is the id for calling.
+	MsgidCall = 0xC0
+
+	// MsgidResponse is the id for responses.
+	MsgidResponse = 0xC1
+
+	// MsgidError is for when there was an error processing the call.
+	MsgidError = 0xC2
+)
+
+type callmsg struct {
+	id     uint32
+	funcID uint16
+	body   []byte
+}
+
+func (m callmsg) Bytes() []byte {
+
+	w := bytes.Buffer{}
+
+	// Write the call ID and message type
+	binary.Write(&w, binary.BigEndian, m.id)
+	binary.Write(&w, binary.BigEndian, m.funcID)
+
+	// Write the body len
+	blen := uint32(len(m.body))
+	binary.Write(&w, binary.BigEndian, blen)
+
+	// Write the body.
+	w.Write(m.body)
+
+	return w.Bytes()
+
+}
+
+func (callmsg) Type() uint8 {
+	return MsgidCall
+}
+
+type respmsg struct {
+	id   uint32
+	body []byte
+}
+
+func (m respmsg) Bytes() []byte {
+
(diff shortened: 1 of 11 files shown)


delbonis:
A potential next step would be to refactor the remote control and message forwarding system to use peer function calls underneath.

delbonis:
retest this please

lnp2p/callmsgs.go
```diff
@@ -188,7 +188,7 @@ func ParseErrMessage(buf []byte) (Message, error) {
 	}
 
 	// More magic to read the error message.
-	var mlen uint32
+	var mlen uint16
```
delbonis: I can't believe I made this mistake the first time around.

lnp2p/peermgr.go
```diff
@@ -449,16 +449,16 @@ func (pm *PeerManager) SetupFunctionalMessageHandling() {
...
 		if !ok {
-			return fmt.Errorf("bad message type")
+			return fmt.Errorf("bad peercall message type (resp)")
 		}
 		return pm.crouter.processReturnResponse(p, rm)
 	})
 
 	// Register error handling.
-	mp.DefineMessage(MsgidResponse, ParseRespMessage, func(p *Peer, m Message) error {
+	mp.DefineMessage(MsgidError, ParseErrMessage, func(p *Peer, m Message) error {
```
delbonis: Also this mistake was annoying too, because I just copy/pasted the above block and didn't think hard enough about it.

delbonis:
Ok so now it all actually works and there's no major issues with it, should write a test using the new `pingpong` peer call.

delbonis:
This kinda gets obsoleted by the new BOLTy stuff, maybe I'll reuse some of this code in the future.

delbonis:
I'm not really sure why we need a "done" category since that project is a long-lived thing, not like how, say, Bitcoin Core uses it for broad but independent changes.

Varunram:
Yeah, I don't know really, I think "Done" is a default state for GitHub automation.

delbonis:
Oh no I added the automaton for it around when we set it up, but I don't know if there's a way to just "delete" cards when closed.

PR #411, opened 2018-10-12
Fixed autoreconnect not working with new lnp2p session code

delbonis:
This'll probably merge conflict with another PR I have open.  But it's just a pretty simple flag thing so it should be easy to fix.

### DIFF
--- a/qln/init.go
+++ b/qln/init.go
@@ -22,7 +22,7 @@ import (
 
 // NewLitNode starts up a lit node.  Needs priv key, and a path.
 // Does not activate a subwallet; do that after init.
-func NewLitNode(privKey *[32]byte, path string, trackerURL string, proxyURL string, nat string) (*LitNode, error) {
+func NewLitNode(privKey *[32]byte, path string, trackerURL string, proxyURL string, nat string, autoreconn bool) (*LitNode, error) {
 
 	var err error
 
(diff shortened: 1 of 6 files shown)


qln/init.go
```diff
@@ -136,6 +136,23 @@ func NewLitNode(privKey *[32]byte, path string, trackerURL string, proxyURL stri
 	nd.PeerMap = map[*lnp2p.Peer]*RemotePeer{}
 	nd.PeerMapMtx = &sync.Mutex{}
 
+	// Do we do autoreconnection things?
+	if autoreconn {
```
gertjaap: Two things need to change here:

1. Don't block the main thread here. Should do this in a goroutine in stead.
2. Using the cached endpoint seems pointless. People can move around, or have a dynamic IP. Plus, the outgoing port will never match the port that lit actually listens on. So the cached endpoint is mostly useless due to NAT. Use the tracker for lookup, if someone is not in the tracker, fail.

gertjaap: https://github.com/gertjaap/lit/commit/5380bfd007711c792ef7dc26112d6ebc639766fa

delbonis:
@gertjaap yeah this PR isn't finished yet, I was trying to get the auto-connection to known peers to work out-of-the-box

qln/init.go
delbonis: Really it should attempt to connect with the known IP address, and if it fails it should do a lookup and try again with the new IP address.  And that should be part of lnp2p, but that's a project for another day.

delbonis:
Going to have to rebase #406 and make some changes to that since it conflicts with some behavior here now.

qln/init.go
delbonis: Fixed.

delbonis:
retest this please

delbonis:
For some reason the default build hangs every time we run it, so let's just leave it as-is.

delbonis:
Just rebased, should be ok now.

PR #414, opened 2018-10-13
Fix bug where remote node can crash lit by issuing an RPC call

Varunram:
Mirrored over from the fix at  gertjaap's memberweekdemo branch.


Varunram:
Small fix, so will open a new PR.

PR #415, opened 2018-10-13
Handle Reorgs correctly

Varunram:
The code that handles reorgs has always been buggy, this somewhat improves the situation. One known bug that I find is when we have small reorgs and the node doesn't give us a whole lot of headers in order for us to trace back history. I'm thinking we should have a timeout and then go with the longest chain, but I'm happy for inputs on this.


delbonis:
We should add some tests for this, probably using `invalidateblock`.

Varunram:
Yeah, I shall solve the headers problem and write a few tests for this. I think there's already a unit test for this, so if its there, I shall edit that.

Varunram:
most cases should be covered now,  tests still left. Please merge #416 before this though, since this contains some commits of that. Alternatively, you could merge this right away by pressing the "Merge commit" button to gran ownership to the commit authors.

uspv/header_test.go
```diff
@@ -1,96 +0,0 @@
-package uspv
```
delbonis: Why was this deleted?

Varunram: this is now part of the python integration test framework

delbonis: But as an integration test, this unit test just checks to see if the USPV code can detect/process reorgs sensibly.  The integration test checks to see if lit functions reasonably when a reorg actually happens.

test/itest_testreorg.py
```diff
@@ -0,0 +1,30 @@
...
+# we produce 9 blocks
+def run_test(env):
+    print("REUNNINSA")
+    bc = env.bitcoind
+    lit1 = env.lits[0]
+    bc.rpc.generate(11)
+    tip = bc.rpc.getblockcount()
+    print("TIP", tip)
+    targetblockhash = bc.rpc.getblockhash(tip - 9)
+    # TODO: Stop and start lit again here
```
delbonis: This isn't `# TODO` anymore is it?

Varunram: yeah, after #420 is merged, will do a rebase and then fix this

uspv/header_test.go
Varunram: Agreed, but `moreWork` is pretty useless unless you have a chain, hence its place in the python framework. Also, its easier to create blocks using bitcoind since you don't need to worry about nonces and stuff whereas if you craft it, you do have to.

test/itest_testreorg.py
delbonis: Also I think this file can just be named "reorg" instead of "testreorg" since that's a little redundant.

uspv/header_test.go
delbonis: But then if lit starts breaking when reorgs happen we won't know if it's because someone refactored the uspv code and broke the something involving reorgs there or if someone wrote code higher up that doesn't sensibly handle reorging.  Sure the blocks we're making in the test aren't really sensible but the important part is that it can calculate which chain has more work, even if the chain is just a few blocks long.

adiabat: There's no cost to keeping this as it should still work, so let's just leave it in.

test/itest_testreorg.py
Varunram: Oh yeah, that makes sense. Will rename and put in the restart code as well.

uspv/header_test.go
Varunram: Will restore.

PR #417, opened 2018-10-14
Defer error for a duplicate outgoing HTLC so exchange multihop payments can complete

metalicjames:
If we already have an outgoing HTLC with a given hash and coin type, we don't want to create another. In the case of an exchange we already have an outgoing HTLC when we receive the payment setup nudge to release the preimage for the payment hash. Therefore we need to defer the error until after we've checked if we know the preimage, because in that case we're not making any new HTLCs anyway.

In the future this needs to be changed to check on a per-channel level, rather than per hash-cointype pair so that routing can be reflexive (pass through the same node multiple times).


delbonis:
@metalicjames Is this ready to merge?

PR #419, opened 2018-10-15
Standardize coinparam HDCoinType allocation

Varunram:
In the future, we can refer to SLIP-044 for main net coin number and then allocate the appropriate number for regtest and testnet coins. The number 10000 was chosen hoping that there wouldn't be more than 10000 coins in existence. This can be increased if desired.

WARNING: If you're already running lit, this will not work with your regtest and mainnet coins. This does not change the behaviour of Vertcoin Mainnet in any way, so those funds are safe.

ref: #340


delbonis:
Changing these values is why I'm in favor of eagerly evaluating HD keys for each channel.  So you create a "channel root key" when you create a new channel and that's kept with the channel for the life of it, and there's (almost) absolutely no way to have keys break outside of changing the root key itself.

treyzania:
Does this update the `REGTEST_COINTYPE` value in testlib.py?

delbonis:
Also, does this update the `REGTEST_COINTYPE` value in testlib.py?

Varunram:
I think the channel root key is a nice idea, dunno about the downsides, so need to ask others about it. Regarding `REGTEST_COINTYPE`, good catch, will update.

Varunram:
retest this please

Varunram:
retest this please

PR #420, opened 2018-10-15
Rebase #384 onto master

Varunram:
https://github.com/mit-dci/lit/pull/384 was a PR that couldn't be added to the other list because of a merge conflict. This is #384 rebased unto current master. @delbonis please verify if this contains all changes. If not, you can push to this fork directly to implement changes.

### DIFF
--- a/test/tests.txt
+++ b/test/tests.txt
@@ -1,6 +1,10 @@
 # Basic tests
 testlib 1
 connect 2
+reconnect 0 simple
+reconnect 0 reordered
+reconnect 0 inbound
+reconnect 0 reordered_inbound
 setgetfee 1
 
 # Tx operations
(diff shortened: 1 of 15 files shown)


delbonis:
If it builds properly then I think it should have all of the important changes.

test/tests.txt
```diff
@@ -10,13 +14,13 @@ send2 2
 
 # Simple channel operations
 fund 2
-close 2 run_test_forward
-close 2 run_test_reverse
+close 2 forward
+close 2 reverse
 # Disabled these to tests until we support breaking right after funding (we don't)
```
delbonis: We can remove this line.

Varunram:
it build fine and has the same diff, but I didn't test it much after that.  Test failure is due to the env not being initiated, so will rebuild.

delbonis:
Should be all ok then.

PR #423, opened 2018-10-17
Fix bug where any remote RPC call crashes lit

Varunram:
Still investigating why peerIds start from 0

ref: #414


delbonis:
I tried to make them start at 0 when I was writing lnp2p code, but it turns out there's some behavior dependent on that (interprets 0 as "no peer") so there's still some work that needs to be done on that.

Varunram:
the proposed fix works.

Issue #425, opened 2018-10-18
Make lit-af verbosity levels the same as lit


Varunram:
So lit has 4 verbosity levels (`nil`, `-v`, `-vv`, `-vvv`), but `lit-af` still follows the old logging conventions (`./lit-af -v 3`). Will put in a PR to fix this.

Varunram:
Closed via 
https://github.com/mit-dci/lit/commit/2ecca10e9d3703052b33e5be5d9f9a8b7fd07359

PR #426, opened 2018-10-18
Improve lit-af command line arg parsing

Varunram:
Used the same `flags` package like we do for `lit` since its more extensible and can host arrays.
ref: #425.


Varunram:
retest this please

PR #427, opened 2018-10-20
Don't call litrpc.Balance each time ls is called

Varunram:
I guess this was a carry over of multiple rebases in the past, so we didn't really catch it.


Varunram:
noob error, one min

delbonis:
That miiiight have been my fault.

PR #429, opened 2018-11-01
Made some changes to peer init in lnp2p

delbonis:
This unifies the setup we do *after* we've established the connection.  I moved the peer init code from the functions for connecting to peers and for listening for new connections to another function so new connections get the same treatment regardless of how it was created.

I also did some changes to the NetSettings struct, it's passed to the peermgr on creation, so we don't have to provide it every time we connect to a peer.  Plus some other minor fixes there.

I also added a `lncore.ParseLnAddr` function that takes a string, verifies that it conforms to what we expect for a lit address, and returns it as a `LnAddr` or returns an error describing why it isn't valid.  I also replaced all of the direct casts to `lncore.LnAddr` across lit to calls to that function.  There's also a `.ToString()` function on it but there's not much interesting about that.

### DIFF
--- a/db/lnbolt/peerdb.go
+++ b/db/lnbolt/peerdb.go
@@ -55,7 +55,13 @@ func (pdb *peerboltdb) GetPeerAddrs() ([]lncore.LnAddr, error) {
 			if k == nil {
 				break
 			}
-			atmp = append(atmp, lncore.LnAddr(string(k)))
+			lnaddr, err := lncore.ParseLnAddr(string(k))
+			if err != nil {
+				logging.Warnf("lnbolt/peerdb: found invalid key in DB as lnaddr: %s (error: %s)", string(k), err.Error())
+				continue
+			}
+
+			atmp = append(atmp, lnaddr)
 		}
 
 		// Now that we have the final array return it.
@@ -126,7 +132,11 @@ func (pdb *peerboltdb) GetPeerInfos() (map[lncore.LnAddr]lncore.PeerInfo, error)
 				return err2
 			}
 
-			ka := lncore.LnAddr(string(k))
+			ka, err2 := lncore.ParseLnAddr(string(k))
+			if err2 != nil {
+				logging.Warnf("lnbolt/peerdb: found invalid key in DB as lnaddr: %s (error: %s)", string(k), err.Error())
+				continue
+			}
 			mtmp[ka] = pi
 
 		}
--- a/lncore/peers.go
+++ b/lncore/peers.go
@@ -1,8 +1,45 @@
 package lncore
 
+import (
+	"fmt"
+
+	"github.com/mit-dci/lit/bech32"
+)
+
 // LnAddr is just a bech32-encoded pubkey.
 // TODO Move this to another package so it's more obviously not *just* IO-related.
 type LnAddr string
+
+// ParseLnAddr will verify that the string passed is a valid LN address, as in
+// ln1pmclh89haeswrw0unf8awuyqeu4t2uell58nea.
+func ParseLnAddr(m string) (LnAddr, error) {
+
+	prefix, raw, err := bech32.Decode(m)
+
+	// Check it's valid bech32.
+	if err != nil {
+		return "", err
+	}
+
+	// Check it has the right prefix.
+	if prefix != "ln" {
+		return "", fmt.Errorf("prefix is not 'ln'")
+	}
+
+	// Check the length of the content bytes is right.
+	if len(raw) > 20 {
+		return "", fmt.Errorf("address too long to be pubkey")
+	}
+
+	return LnAddr(m), nil // should be the only place we cast to this type
+
+}
+
+// ToString returns the LnAddr as a string.  Right now it just unwraps it but it
+// might do something more eventually.
+func (lnaddr LnAddr) ToString() string {
+	return string(lnaddr)
+}
 
 // LitPeerStorage is storage for peer data.
 type LitPeerStorage interface {
--- a/lnp2p/listen.go
+++ b/lnp2p/listen.go
@@ -2,7 +2,6 @@ package lnp2p
 
 import (
 	"github.com/mit-dci/lit/eventbus"
-	"github.com/mit-dci/lit/lncore"
 	"github.com/mit-dci/lit/lndc"
 	"github.com/mit-dci/lit/logging"
 )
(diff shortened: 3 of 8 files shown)


lncore/peers.go
```diff
@@ -1,9 +1,46 @@
...
+		return "", fmt.Errorf("address too long to be pubkey")
+	}
+
+	return LnAddr(m), nil // should be the only place we cast to this type
+
+}
+
+// ToString returns the LnAddr as a string.  Right now it just unwraps it but it
+// might do something more eventually.
+func (lnaddr LnAddr) ToString() string {
```
Varunram: This typecast is redundant, `LnAddr` is a string right now, we're really just converting from one name to another

lncore/peers.go
```diff
@@ -1,9 +1,46 @@
...
+	if prefix != "ln" {
+		return "", fmt.Errorf("prefix is not 'ln'")
+	}
+
+	// Check the length of the content bytes is right.
+	if len(raw) > 20 {
+		return "", fmt.Errorf("address too long to be pubkey")
+	}
+
+	return LnAddr(m), nil // should be the only place we cast to this type
```
Varunram: As commented below, this is redundant since we can just return `m`

lncore/peers.go
```diff
@@ -1,9 +1,46 @@
...
 
+import (
+	"fmt"
+
+	"github.com/mit-dci/lit/bech32"
+)
+
 // LnAddr is just a bech32-encoded pubkey.
 // TODO Move this to another package so it's more obviously not *just* IO-related.
 type LnAddr string
```
Varunram: This type definition should really be thought about. My suggestion is that we keep this as string and remove the typecast for now until we find purpose for it. Unless you have other plans :)

db/lnbolt/peerdb.go
```diff
@@ -126,7 +132,11 @@ func (pdb *peerboltdb) GetPeerInfos() (map[lncore.LnAddr]lncore.PeerInfo, error)
 				return err2
 			}
 
-			ka := lncore.LnAddr(string(k))
+			ka, err2 := lncore.ParseLnAddr(string(k))
```
Varunram: `ka`? (also nit, but could be just `err`)

db/lnbolt/peerdb.go
```diff
@@ -55,7 +55,13 @@ func (pdb *peerboltdb) GetPeerAddrs() ([]lncore.LnAddr, error) {
 			if k == nil {
 				break
 			}
-			atmp = append(atmp, lncore.LnAddr(string(k)))
+			lnaddr, err := lncore.ParseLnAddr(string(k))
+			if err != nil {
+				logging.Warnf("lnbolt/peerdb: found invalid key in DB as lnaddr: %s (error: %s)", string(k), err.Error())
```
Varunram: shouldn't we ideally check while writing and not reading? one should be enough if we're doing it right.

lnp2p/listen.go
```diff
@@ -48,58 +47,23 @@ func acceptConnections(listener *lndc.Listener, port int, pm *PeerManager) {
...
 
-		logging.Infof("New connection from %s at %s\n", rlitaddr, rnetaddr.String())
-
-		// Read the peer info from the DB.
-		pi, err := pm.peerdb.GetPeerInfo(rlitaddr)
-		if err != nil {
-			logging.Warnf("problem loading peer info in DB (maybe this is ok?): %s\n", err.Error())
-			netConn.Close()
+		// Make sure we can't let ourself connect to ourself.
+		if string(rlitaddr) == pm.GetExternalAddress() {
```
Varunram: Nice catch!

lnp2p/peermgr.go
```diff
@@ -353,9 +377,8 @@ func (pm *PeerManager) ListenOnPort(port int) error {
 		return fmt.Errorf("listen cancelled by event handler")
 	}
 
-	// TODO UPnP and PMP NAT traversal.
-
 	// Try to start listening.
+	// TODO Listen on proxy if possible?
```
Varunram: Proxy on localhost?

qln/init.go
```diff
@@ -60,7 +60,7 @@ func NewLitNode(privKey *[32]byte, path string, trackerURL string, proxyURL stri
 	nd.Events = &ebus
 
 	// Peer manager
-	nd.PeerMan, err = lnp2p.NewPeerManager(rootPrivKey, nd.NewLitDB.GetPeerDB(), trackerURL, &ebus)
+	nd.PeerMan, err = lnp2p.NewPeerManager(rootPrivKey, nd.NewLitDB.GetPeerDB(), trackerURL, &ebus, nil) // TODO proxy/nat stuff
```
Varunram: I think NAT stuff is moved? Didn't test this myself, but I see that you moved it.

lnp2p/peermgr.go
```diff
@@ -156,81 +156,81 @@ func (pm *PeerManager) TryConnectAddress(addr string, settings *NetSettings) (*P
...
-		conn:     lndcconn,
+		conn:     conn,
 		idpubkey: pk,
+
+		// TEMP
+		idx: nil,
+	}
+
+	pi, err := pm.peerdb.GetPeerInfo(expectedAddr)
+	if err != nil {
```
Varunram: This check could actually be `if err != nil || pi == nil {`

Varunram: Also, if we fall through to `logging.Errorf("Problem getting unique peeridx from DB: %s\n", err.Error())`, could we assign a random number above 10000 or something? Because we don't catch that and it could result in some weird case where a peer doesn't have a peerId.

Varunram:
Also, would be nice if you could create a new label named "In Progress" to tag it so :)

db/lnbolt/peerdb.go
delbonis: "`k` as `a`ddress"

And `go vet` was complaining something about shadowing so I made it a different name.

lncore/peers.go
delbonis: This call *is* redundant but it makes it clear that you shouldn't treat a lnaddr as a plain string.  The compiler should inline it out altogether.

lncore/peers.go
delbonis: The point of it being a separate type is so that you have the guarantee that it's a properly-formatted ln address.  It also changes the way you think about processing addresses so that it's though of more like how you'd parse binary data into a structure.

Of course in Go this is a weak guarantee, but if it was tracked as a `string` then it could lead to problems when refactoring (and accidentally removing checks) that lead to random errors.  That's why I removed all the other force casts to/from the `LnAddr` type.  With this you have some typing guarantee that you won't (can't) forget to verify anything.

qln/init.go
delbonis: I used it incorrectly the first time, but now the issue left is that we don't have plumbing set up to take it from the args yet.

db/lnbolt/peerdb.go
delbonis: Benefits outweigh the risks, if "magic" happens and the DB gets corrupted then it would be good to have an error about it here before it caused problems somewhere else where the original source would have been unknown.

lnp2p/peermgr.go
delbonis: My goal is to eventually not have peer IDs externally and at most use them ephemerally within the process.

I guess that works though?  I don't really see how it solves the problem though.

db/lnbolt/peerdb.go
Varunram: oh, `go vet` stuff, cool

db/lnbolt/peerdb.go
Varunram: a partial corruption where the key exists but the value doesn't? seems a bit weird, but yeah, doesn't hurt to have one I guess.

lnp2p/peermgr.go
Varunram: We could do way with peer IDs, but that would remove of a wait to easily reference them by the end user. I guess they can do something similar with nicknames, but we must display that in `lit-af`'s output.

It doesn't solve the problem (which in this case is a peer not having a peerId), but circumvents it by assigning a random peerId. Given that the case where not having a peerId is rare (and maybe never occurs in practice), maybe we could just skip it totally.

delbonis: RPCs for BOLT-family LN imlplementations just use the peer's full address, which imo makes sense.  And since our lnaddrs are shorter it's more readable anyways too.  And nicknames would make aliasing that simpler as well, perhaps we should allow the users to provide a nickname when initially connecting.

Varunram: agreed.

PR #434, opened 2018-11-21
Accept pubkey connections in lit

Varunram:
Instead of just accepting raw pk hashes, accept pubkeys as well so that we can connect using them. A small precursor to the bigger pubkey revamp that's coming. Requires #432 and builds upon one of its commit.


Varunram:
Closed in favour of #437

PR #437, opened 2018-11-23
Connect and listen to BOLT compatible nodes

Varunram:
Rebased on current master


delbonis:
Should those other PRs be closed now?

cmd/lit-af/lit-af.go
```diff
@@ -163,6 +163,13 @@ func (lc *litAfClient) litAfSetup(conf litAfConfig) error {
 		if err != nil {
 			logging.Fatal(err.Error())
 		}
+	} else if len(adr) == 0 {
+		// so the user didn't provide us with an address to connect to
+		// we need to connect to the locally running lit-af instance
+		lc.RPCClient, err = litrpc.NewLocalLndcRpcClientWithHomeDirAndPort(defaultDir, port)
```
delbonis: Can we make a way to specify a homedir too?  Because without that it kinda breaks my workflow.

Varunram: hm.. yeah, I think that would be possible, was possible earlier as well iirc. Will push an update.

Varunram: This seems to work right now (`./lit-af --dir=blah`), was this the use case you were looking at?

Varunram:
retest this please

Varunram:
Closed other PRs, squashed the commits in this one.

Varunram:
retest this please

PR #438, opened 2018-11-23
Resolve bug where reorging greater than attachheight requires db wipe

Varunram:


Varunram:
the tests which are in place now work, don't know why / how the older test failed.

PR #440, opened 2018-11-27
New more flexible address format code

delbonis:
Adds partial support for decoding both Lit-style and BOLT-style LN addresses.

See `lncore/addr.go`.

### DIFF
--- a/lncore/addr.go
+++ b/lncore/addr.go
@@ -1,0 +1,145 @@
+package lncore;
+
+import (
+	"fmt"
+	
+	"github.com/mit-dci/lit/bech32"
+)
+
+/*
+ * Ok so there's a lot of stuff going on here.  Lit natively supports addresses
+ * that are just the pkh of the pubkey encoded in bech32.  And that's pretty
+ * cool, you look up the IP+port in a tracker (or DHT) and connected with
+ * Noise_XX, then verify that the pubkey they provide you matches the pkh you
+ * already have.  Also you can provide IP+port if you want.
+ *
+ * But in mainstream BOLT nodes they use the full pubkey hex-encoded.  They have
+ * a DHT where you look up the IP+port (if not provided) and connect with
+ * Noise_XK, which assumes you already have the pubkey (since you do) and that's
+ * all fine.
+ *
+ * This tries to unify all of them with formats like this under 1 parser:
+ *
+ *      <bech32 pkh>[@<ip>[:<port>]
+ * or
+ * 
+ *      <hex of bech32 pubkey>[@<ip>[:<port>]
+ *
+ * or fully
+ *
+ *      <bech32 pkh>:<pubkey>[@<ip>:<port>]
+ */
+
+// LnDefaultPort is from BOLT1.
+const LnDefaultPort = 9735
+
+// PkhBech32Prefix if for addresses lit already uses.
+const PkhBech32Prefix = "ln"
+
+// PubkeyBech32Prefix is for encoding the full pubkey in bech32.
+const PubkeyBech32Prefix = "lnpk"
+
+// LnAddressData is all of the data that can be encoded in a parsed address.
+type LnAddressData struct {
+	Pkh    *string
+	Pubkey []byte
+	IPAddr *string
+	Port   uint16
+}
+
+// AddrFmtFull is <bech32 pkh>:<hex pk>@<ip>:<port>
+const AddrFmtFull = "full"
+
+// AddrFmtFullBech32Pk is <bech32 pkh>:<bech32 pk>@<ip>:<port>
+const AddrFmtFullBech32Pk = "full_bech32_pk"
+
+// AddrFmtFullNoPort is <bech32 pkh>:<hex pk>@<ip>
+const AddrFmtFullNoPort = "full_no_port"
+
+// AddrFmtFullBech32PkNoPort is <bech32 pkh>:<bech32 pk>@<ip>
+const AddrFmtFullBech32PkNoPort = "full_bech32_pk_no_port"
+
+// AddrFmtBech32 is just <bech32 pkh>
+const AddrFmtBech32 = "bech32_pkh"
+
+// AddrFmtLit is just AddrFmtBech32
+const AddrFmtLit = "lit"
+
+// AddrFmtPubkey is <hex pubkey>
+const AddrFmtPubkey = "hex_pk"
+
+// AddrFmtBech32Pubkey is <bech32 pk>
+const AddrFmtBech32Pubkey = "bech32_pk"
+
+// AddrFmtPubkeyIP is <hex pk>@<ip>
+const AddrFmtPubkeyIP = "hex_pk_ip"
+
+// AddrFmtPubkeyIPPort is <hex pk>@<ip>:<port>
+const AddrFmtPubkeyIPPort = "hex_pk_ip_port"
+
+// AddrFmtLitFull is <bech32 pkh>@<ip>:<port>
+const AddrFmtLitFull = "lit_full"
+
+// AddrFmtLitIP is <bech32 pkh>@<ip>
+const AddrFmtLitIP = "lit_ip"
+
+// AddrDataToFormats returns the addresses in all of the formats that fully represent it.
+func DumpAddressFormats(data LnAddressData) (map[string]string, error) {
+
+	ret := make(map[string]string)
+
+	if data.Port != LnDefaultPort && data.IPAddr != nil {
+		return ret, fmt.Errorf("nondefault port specified but IP not specified")
+	}
+
+	// Full.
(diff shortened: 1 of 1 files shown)


delbonis:
Not that lnp2p doesn't use this code at all yet, we should have another PR that actually uses this address parser for figuring out connections.

lncore/addr.go
```diff
@@ -0,0 +1,319 @@
...
+
+	"github.com/mit-dci/lit/bech32"
+	"github.com/mit-dci/lit/crypto/fastsha256"
+)
+
+/*
+ * Ok so there's a lot of stuff going on here.  Lit natively supports addresses
+ * that are just the pkh of the pubkey encoded in bech32.  And that's pretty
+ * cool, you look up the IP+port in a tracker (or DHT) and connected with
+ * Noise_XX, then verify that the pubkey they provide you matches the pkh you
```
Varunram: So how lit should work after #437 is that if we haven't connected to the peer before, we still go ahead with NOISE_XX and get the pubkey of the host. But if we are connecting to a host whose pubkey we know already, we use NOISE_XK, since we already know the remote node's pubkey and theres no requirement to ask again. This is similar to how we using `ssh` asks you if you want to remember and I guess an implementing UI could ask the user for his preference as well

lncore/addr.go
```diff
@@ -0,0 +1,319 @@
...
+type LnAddr string
+
+// ToString returns the LnAddr as a string.  Right now it just unwraps it but it
+// might do something more eventually.
+func (lnaddr LnAddr) ToString() string {
+	return string(lnaddr)
+}
+
+// LnDefaultPort is from BOLT1.
+const LnDefaultPort = 9735
```
Varunram: we could rename this to `BOLTDefaultPort` or just plain `DefaultPort` since Ln is implied :)

lncore/addr.go
```diff
@@ -0,0 +1,319 @@
...
+
+// PkhBech32Prefix if for addresses lit already uses.
+const PkhBech32Prefix = "ln"
```
Varunram: Another small issue, but this is just `lit`'s bech32 prefix and it would be better named that way

lncore/addr.go
```diff
@@ -0,0 +1,319 @@
...
+
+// PubkeyBech32Prefix is for encoding the full pubkey in bech32.
+const PubkeyBech32Prefix = "lnpk"
```
Varunram: Why would we need this though, can't we parse with just the length of the resultant `bech32` string?

lncore/addr.go
```diff
@@ -0,0 +1,319 @@
...
+// checks if there's an IP address defined.
+func (d *LnAddressData) HasFullAddress() bool {
+	return d.IPAddr != nil
+}
+
+// AddrFmtFull is <bech32 pkh>:<hex pk>@<ip>:<port>
+const AddrFmtFull = "full"
+
+// AddrFmtFullBech32Pk is <bech32 pkh>:<bech32 pk>@<ip>:<port>
+const AddrFmtFullBech32Pk = "full_bech32_pk"
```
Varunram:
```
// AddrFmtFullBech32Pk is <bech32 pkh>:<bech32 pk>@<ip>:<port>
```
in the above, if we have the bech32 pk, we could easily derive the bech32 kph (`LitAddr`). Maybe we could just have the bech32 pk in full and then use the relevant methods where we need the pkh?

lncore/addr.go
```diff
@@ -0,0 +1,319 @@
...
+const AddrFmtFullBech32PkNoPort = "full_bech32_pk_no_port"
+
+// AddrFmtBech32 is just <bech32 pkh>
+const AddrFmtBech32 = "bech32_pkh"
+
+// AddrFmtLit is just AddrFmtBech32
+const AddrFmtLit = "lit"
+
+// AddrFmtPubkey is <hex pubkey>
+const AddrFmtPubkey = "hex_pk"
```
Varunram: Again, unless we're using this in some internal parsing, the hex_pk on its own is not that useful (since there's no tracker that `lnd` uses). How do you envision using this?

lncore/addr.go
```diff
@@ -0,0 +1,319 @@
...
+
+// AddrFmtBech32Pubkey is <bech32 pk>
+const AddrFmtBech32Pubkey = "bech32_pk"
```
Varunram: Somewhat similar to the comment above in that the bech32 pk can be converted to a bech32 pkh and then used instead of parsing pks and pkh's separately. I assume that requiring the full bech32 pk would not be needed at all (except for internal parsing, where we could use  defined methods / functions)

lncore/addr.go
```diff
@@ -0,0 +1,319 @@
...
+		ret[AddrFmtPubkey] = fmt.Sprintf("%X", data.Pubkey)
+		if data.IPAddr != nil {
+			if data.Port == LnDefaultPort {
+				ret[AddrFmtPubkeyIP] = fmt.Sprintf("%X@%s", data.Pubkey, *data.IPAddr)
+			}
+			ret[AddrFmtPubkeyIPPort] = fmt.Sprintf("%X@%s:%d", data.Pubkey, *data.IPAddr, data.Port)
+		}
+	}
+
+	// TODO more
```
Varunram: I think these formats are enough actually, we just need two formats (maybe three, if you count just the pkh as one format).
1. Lit Addr (pkh) with Port
2. BOLT Addr (hex pubkey) with Port
(3). Lit Addr (pkh)

Again, dependent on how we would use this since displaying addresses requires only the two formats above unless I'mm missing something here.

lncore/addr.go
```diff
@@ -0,0 +1,319 @@
...
+	
+	return addrdata, nil // TODO
+}
+
+/*
++0      q 	p 	z 	r 	y 	9 	x 	8
++8 	g 	f 	2 	t 	v 	d 	w 	0
++16 	s 	3 	j 	n 	5 	4 	k 	h
++24 	c 	e 	6 	m 	u 	a 	7 	l
+*/
```
Varunram: I guess we could omit this? (Or add this as a comment below the bech32 regex).

lncore/addr.go
delbonis: Yeah that's true, I suppose I wanted to make the difference more obvious.  Especially since technically `ln1dfasdfasdf:lnpk1asdfsadfasdf` is a valid address, it'd look a little weird to me if it was `ln1:ln1`, and it'd be less obvious for the average joe to know how to pick out the "short" version of their address, which we want to make the pkh.

lncore/addr.go
delbonis: RPC though?  Although that *is* in a different package so maybe it doesn't matter.

lncore/addr.go
delbonis: In c-lightning you can connect to nodes if you *just* have the pubkey, I'm not sure how it looks up the IP address but it *is* a valid "thing" you can connect with.

lncore/addr.go
delbonis: No it's not really necessary but I thought I'd include it "just because".

lncore/addr.go
delbonis: Yeah that was to help explain the regex.

lncore/addr.go
delbonis: Yeah people would probably only end up using those 3 formats often, but since people do weird things I thought it makes sense to include all possible variations?  Especially since it's not hard for a person to manually convert most formats to simpler versions.

lncore/addr.go
delbonis: Oh yes you can definitely derive the pkh from the pk.  I hope nobody uses this format except when making fully-qualified addresses for whatever reason.  But in practical use I don't really see why anyone would use it.  I left it (as with the variant with the hex pk) just as "one option".

lncore/addr.go
delbonis: Fixed, along with the port const naming.

lncore/addr.go
delbonis: Fixed.

lncore/addr.go
Varunram: Sounds good thanks! Although I wonder why people would want to use bech32 pks (since lit right now supports only bech32 pkh and BOLT supports hex pks). Would be nice if some implementations decide to change though, so having this does make sense.

lncore/addr.go
Varunram: Yeah, I guess with your overall rationale in the other comments, I can see why you would want to include other formats here.

lncore/addr.go
delbonis: I mean, it doesn't *really* matter.  They get a little bit of error correction (maybe), but since most of the time addresses are going to be copy/pasted I don't think it matters.  I added it to quell arguments about "well hex might be bad but bech32 is better" since this is demonstrably not that much shorter and not worth the extra effort.

PR #441, opened 2018-11-30
Cleaned up some redundancies in the lit-af ls command

### LINKED ISSUE #433: ls conns output shows channels
when we do `ls conns`, we should ideally show only the active connections (and not `Channels`)
```
entered command: ls conns
1 127.0.0.1:53627 (ln1qy95q25hs7t43779gqqetswdefrx6ze2f02ajz)
	Channels:
```

delbonis:
This should resolve #433.

### DIFF
--- a/cmd/lit-af/shell.go
+++ b/cmd/lit-af/shell.go
@@ -343,22 +343,9 @@ func (lc *litAfClient) Ls(textArgs []string) error {
 		}
 	}
 
-	err = lc.Call("LitRPC.Balance", nil, bReply)
-	if err != nil {
-		return err
-	}
-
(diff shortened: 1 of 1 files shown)


cmd/lit-af/shell.go
```diff
@@ -25,7 +25,7 @@ var lsCommand = &Command{
 			"What information to show. Provide one of:"),
 		fmt.Sprintf("\t%-20s %s",
 			lnutil.White("-a"),
-			"Information on connections to other peers"),
+			"Everything"),
```
Varunram: Nice catch :)

Varunram:
retest this please

Issue #449, opened 2019-03-08
Discrete log contracts and Noise Protocol message size limitation is 64kb.


infografx:
The issue is in that when I setup the contract as in 
the documentation:

```
dlc contract setfunding 1 100000000 100000000
dlc contract setdivision 1 20000 10000
```
the commands:

```
dlc contract ls
dlc contracts view 1
```

are not working because the max size of a go channel item is 64kb.

It just can't send all
```
 }
   "OracleValue": 19997,
   "ValueOurs": 199940000
  },
  {
   "OracleValue": 19998,
   "ValueOurs": 199960000
  }
```
values throug the channel.

====================================

The question is:

Can we change the contact so that the variables "OracleValue" and  "ValueOurs" are calculated dynamically.

Then create RPC call to retrive "ValueOurs", for example:

```
lit-af: dlc contract 1 ours 19997
```

Where 19997 is the "OracleValue".

And it will show corresponding ValueOurs:

```
lit-af: 
Command entered: dlc contract 1 ours 19997
"ValueOurs": 199940000
```

But this value have to be calculated on the client side, based on variables from

lit/dlc/contract.go:  SetContractDivision function:
```
rangeMin
rangeMax
valueAllOurs
valueAllTheirs
```
Thanks.

infografx:
So what do you think about this.

I want to complete it but it would be great if you comment it first.

infografx:
Ok. Maybe the answer is in whitepaper:
https://adiabat.github.io/dlc.pdf

delbonis:
Could possibly just pass the pointer to the struct in the channel instead of the entire struct.

infografx:
Hello guys.

I made a pointer, after that it became clear that the problem was as follows:

the file: lit/lndc/conn.go

```the function: func (c *Conn) Write(b []byte) (n int, err error)```
This function chunks large messages.

```the function: func (c *Conn) Read(b []byte) (n int, err error)```
But this function just read them. Without handling chunks.
This causes messages larger than 64 kilobytes to be processed incorrectly.

--------

Interestingly, in the LND node code, in this place is the same.

Thus, or it it has the same problem, or I do not understand something.

---------

PROPOSAL.

Wrap each message to BEGIN, and END messages.

BEGIN message format:


type RemoteControlRpcBeginMsg struct {
	PeerIdx uint32
	Idx     uint64 // Unique nonce that was sent in the request
	ChNum   bool   // Number of chunks
}

END message format (the same):

[...]

PR #451, opened 2019-03-21
Tests for the DLC subsystem.

infografx:
Also I have wrapped all test functions in try/except to avoid resources leak(bitcoind/dlcoracle/lit) when some RPC calls get broken, or something else getting wrong.

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,7 +77,13 @@ The below command will run Lit on the Bitcoin testnet3 network
 
 The words `yup, yes, y, true, 1, ok, enable, on` can be used to specify that Lit
 automatically connect to peers fetched from a list of DNS seeds. It can also be replaced by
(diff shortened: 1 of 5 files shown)


infografx:
So what do you think about:
https://github.com/mit-dci/lit/issues/449

I want to complete it but it would be great if you comment it first.

delbonis:
I probably would have made 2 separate PRs to do the thing with the `try-catch` and then add the DLC tests.

Also if you wanted to be through you should make multiple DLC tests:

* opening contracts
* resolving contracts with oracle sig
* claiming timed out contracts without oracle sig
* handling oracle mischief

infografx:
Ok. For now I see that only one node receive it's amount of the money after a contract settled.
If, in your case, both nodes get their money, let me know.

PR #453, opened 2019-04-19
Arbitrary message size

infografx:
I wrap chunks of large messages in to another messages.


infografx:
Now large DLCs are handled correctly.

infografx:
But I have to admit that it’s difficult to understand this PR because major changes are confused with minor ones.

infografx:
This PR is to avoid Noise Protocol message size limitation (64kb).
https://noiseprotocol.org/noise.html
(All Noise messages are less than or equal to 65535 bytes in length.)

I think that even with the optimization by 'Base and Exponent R values'
a contract can be larger than 64kb.

infografx:
Hello.
I'm still working on the dlc tests.
The tests above, of course cannot test updated messaging subsystem.

PR #454, opened 2019-04-22
Implement parts of the event system for channels, update utils for coinparam

Rjected:
* Changed all of the occurrences where `btcutil/chaincfg` is used to our own `coinparam`. While doing this I added `ForkID` to the `Params` struct, since this may be needed to support certain forks that use a Fork ID. I then deleted the `chaincfg/params.go` file because it's no longer used or needed since coinparam is a superset of the chaincfg params. 

 * Added a go.mod and go.sum to support Go 1.11 modules.

 * There was a segfault that happened in `uspv/header.go` since an index was not being checked, so I made that return an error instead.

 * Implemented some event publishing and modified the channel state update events so they publish relevant information.

 * Fixed an issue where channels could not be created with 0 on the other side of the channel.

 * Added two functions to the `ChainHook` interface:
```go
	// NewRawBlocksChannel returns a new channel to receive blocks.
	NewRawBlocksChannel() chan *wire.MsgBlock

	// NewHeightChannel returns a new channel to receive height events.
	NewHeightChannel() chan int32

```
I added this because while working on opencx I was running two chainhooks, one that ran in lit already, and another to detect confirmations. This is just so there's no redundancy when it comes to listening for chain events. So if I'd like to use the same chainhook for multiple things I can, and the channels work pretty much the same as `RawBlocks()`.

 * Just ran `gofmt ./...` from the root directory


infografx:
Previous PR https://github.com/mit-dci/lit/pull/453 has not conflicts with this PR.

infografx:
But I have to admit that it’s difficult to understand this PR because major changes are confused with minor ones. :)

adiabat:
Looks good, it's much cleaner having just coinparams instead of remaining chaincfg.

PR #457, opened 2019-07-22
Arbitrary message size

infografx:
This PR is to avoid Noise Protocol message size limitation (64kb).
https://noiseprotocol.org/noise.html
(All Noise messages are less than or equal to 65535 bytes in length.)

I think that even with the optimization by 'Base and Exponent R values'
a contract can be larger than 64kb.


infografx:
If you have any questions. You are welcome.

infografx:
Or if you have comments or suggestions. Welcome.

PR #459, opened 2019-07-22
DLC Subsystem. Contract settlement from counterparty works.

infografx:
This allows you to run a contract from another node. Not from that which offers the contract.


infografx:
If you have any questions. You are welcome.

infografx:
Or if you have comments or suggestions. Welcome.

PR #467, opened 2019-10-27
Fix makefile

Rjected:
Because `jackpal/go-nat-pmp` hasn't had a release in years and we use `natpmp.NewClientWithTimeout` (which is not defined in the latest release, only in master), running `make` results in this:
```
dan@dan-pc ~/D/P/lit> make
build/env.sh go get -v ./...
go: github.com/jackpal/go-nat-pmp@v0.0.0-20181021122511-d89d09f6f332: invalid pseudo-version: does not match version-control timestamp (2018-10-21T19:25:11Z)
make: *** [Makefile:11: goget] Error 1
```
Interestingly enough, `go build` does not necessarily return the same error.
So I just updated the timestamp, and now the makefile works.


Rjected:
#469 is better than this

PR #470, opened 2019-11-19
fix invalid pseudo-version of go-nat-pmp

forchain:
20181021122511 is a local time (time zone -7) , so it will break my building using "go build"

go: github.com/jackpal/go-nat-pmp@v0.0.0-20181021122511-d89d09f6f332: invalid pseudo-version: does not match version-control timestamp (2018-10-21T19:25:11Z)


Rjected:
#469 fixes this I'm pretty sure

forchain:
@Rjected Hi, I saw your commit. But the comment was saying not to update the version to the latest in case of some issue, so I  just fixed the timestamp but kept the revision untouched . Anyway, I will close it first, if your commit rejected, I will reactivate it.

